PR #61, opened 2015-06-18
[v2] build system fixes, cleanups, and s390 support

michich:
Hello Avner,

I modified my patchset according to your review comments.
The differences from my previous pull request are:
- I added an improved method of extending CXXFLAGS and macros to check if compiler accepts the requested flags. By default "-O3" and "--param inline-unit-growth=200" are still applied.
- I added a macro to unify the checking for boolean ./configure options. Options like "--enable-doc" and "--enable-doc=yes" work, and "--enable-doc=blabla" is correctly rejected.
- The "--enable-build32" options is preserved.
- The unused _DEBUG macro is gone.


avnerbh:
Hi Michal,
Thanks for work and for all your commits.
My Apologizes for the time it took me to handle it.
I appreciate your contribution.

Cheers,
  Avner

PR #62, opened 2015-09-21
workarounds for RHEL6 autoconf

michich:
Hello Avner,
this works around a couple of bugs in old autoconf versions, such as the version in RHEL6.


avnerbh:
Thanks Michal. I apperciate that.

PR #67, opened 2016-03-30
test: Fix verifier

igor-ivanov:


igor-ivanov:
@avnerbh please look at and accept

PR #72, opened 2016-06-14
update Feed File mode (IOMux) to use epoll by default

rosenbaumalex:
epoll performance is better then select/poll so we want to use it by default.

Signed-off-by: Alex Rosenbaum alexr@mellanox.com

### DIFF
--- a/doc/main.dox
+++ b/doc/main.dox
@@ -116,10 +116,10 @@ sockperf: [tid 4857] using select() to block on socket(s)
  -i      --ip                 	Listen on/send to ip <ip>.
  -p      --port               	Listen on/connect to port <port> (default 11111).
  -m      --msg-size           	Use messages of size <size> bytes (minimum default 14).
- -f      --file               	Tread multiple ip+port combinations from file <file> 
-                                (server uses select).
+ -f      --file               	Read multiple ip+port combinations from file <file> 
+                                (will use IO muxer '-F' such as epoll, poll or select)
  -F      --io-hanlder-type    	Type of multiple file descriptors handle 
-                                [s|select|p|poll|e|epoll](default select).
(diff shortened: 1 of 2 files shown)


doc/main.dox
```diff
@@ -96,7 +96,7 @@ sockperf: [SERVER] listen on:
 [ 2] IP = 192.168.161.129 PORT =  7003 # TCP
 [ 3] IP = 192.168.161.129 PORT =  7004 # TCP
 sockperf: Warmup stage (sending a few dummy packets)...
-sockperf: [tid 4857] using select() to block on socket(s)
+sockperf: [tid 4857] using epoll() to block on socket(s)
```
avnerbh: this edit is a mistake, because the command line above it says: 
  $bin/sockperf server -f conf.file **-F s**

doc/main.dox
```diff
@@ -117,9 +117,9 @@ sockperf: [tid 4857] using select() to block on socket(s)
  -p      --port               	Listen on/connect to port <port> (default 11111).
  -m      --msg-size           	Use messages of size <size> bytes (minimum default 14).
  -f      --file               	Tread multiple ip+port combinations from file <file> 
-                                (server uses select).
+                                (server uses epoll).
```
avnerbh: you may consider something general:
(server uses epoll) -> will use IO muxer such as select, poll, or epoll

BTW, do you know what is "Tread" here?

rosenbaumalex: "Tread" looks like a typo... I'll fix to "Read"

I'll also use your suggested working: "will use IO muxer such as select, poll, or epoll"

avnerbh: All looks good to me

PR #78, opened 2016-11-20
Add --use-dummy parameter to support VMA's Dummy send feature.

liranoz12:
WARNING - Using this parameter without VMA will cause all packets to be
sent.

### DIFF
--- a/src/Client.h
+++ b/src/Client.h
@@ -313,7 +313,7 @@ private:
 		m_switchMsgSize.execute(m_pMsgRequest);
 
 		//idle
-		m_switchCycleDuration.execute(ifd, m_pMsgRequest->getLength());
+		m_switchCycleDuration.execute(m_pMsgRequest, ifd);
 
 		//send
 		for (unsigned i = 0; i < g_pApp->m_const_params.burst_size && !g_b_exit; i++) {
--- a/src/Client.h
+++ b/src/Client.h
@@ -313,7 +313,7 @@ private:
 		m_switchMsgSize.execute(m_pMsgRequest);
 
 		//idle
-		m_switchCycleDuration.execute();
+		m_switchCycleDuration.execute(ifd, m_pMsgRequest->getLength());
 
 		//send
 		for (unsigned i = 0; i < g_pApp->m_const_params.burst_size && !g_b_exit; i++) {
--- a/src/Defs.h
+++ b/src/Defs.h
@@ -187,7 +187,8 @@ enum {
 	OPT_CLIENTIP,                   //37
 	OPT_TOS,                        //38
 	OPT_LLS,                        //39
-	OPT_MC_SOURCE_IP                //40
+	OPT_MC_SOURCE_IP,               //40
+	OPT_DUMMY_SEND                  //41
 };
 
 #define MODULE_NAME			"sockperf"
@@ -603,6 +604,7 @@ struct user_params_t {
 	int tos;
 	unsigned int lls_usecs;
 	bool lls_is_set;
+	bool b_dummy_send; //client side only
 };
 
 struct mutable_params_t {
--- a/src/Defs.h
+++ b/src/Defs.h
@@ -147,6 +147,8 @@ extern const int MAX_FDS_NUM;
 #define DEFAULT_SELECT_TIMEOUT_MSEC	10
 #define DEFAULT_DEBUG_LEVEL			0
 
+#define DUMMY_SEND_FLAG MSG_SYN // Used by offload libraries to do egress path warm-up of caches. It is not in use by kernel. WARNING: it will actually end this packet on the wire.
+#define DUMMY_SEND_MPS_DEFAULT 10000
 
 enum {
 	OPT_RX_MC_IF 			 = 1,
@@ -604,7 +606,9 @@ struct user_params_t {
 	int tos;
 	unsigned int lls_usecs;
 	bool lls_is_set;
-	bool b_dummy_send; //client side only
+	bool dummy_send; //client side only
+	uint32_t dummy_mps; //client side only
+	TicksDuration dummySendCycleDuration; //client side only
 };
 
 struct mutable_params_t {
--- a/src/Defs.h
+++ b/src/Defs.h
@@ -150,7 +150,7 @@ extern const int MAX_FDS_NUM;
 /*
 Used by offload libraries to do egress path warm-up of caches.
 It is not in use by kernel. WARNING: it will actually end this packet on the wire.
-More info at vma_extra.h
+DUMMY_SEND_FLAG value should be compatible with the value of VMA_SND_FLAGS_DUMMY (More info at vma_extra.h).
 */
 #define DUMMY_SEND_FLAG MSG_SYN // equals to 0x400
 #define DUMMY_SEND_MPS_DEFAULT 10000
--- a/src/Defs.h
+++ b/src/Defs.h
@@ -147,7 +147,12 @@ extern const int MAX_FDS_NUM;
 #define DEFAULT_SELECT_TIMEOUT_MSEC	10
 #define DEFAULT_DEBUG_LEVEL			0
 
-#define DUMMY_SEND_FLAG MSG_SYN // Used by offload libraries to do egress path warm-up of caches. It is not in use by kernel. WARNING: it will actually end this packet on the wire.
+/*
+Used by offload libraries to do egress path warm-up of caches.
+It is not in use by kernel. WARNING: it will actually end this packet on the wire.
+More info at vma_extra.h
+*/
+#define DUMMY_SEND_FLAG MSG_SYN // equals to 0x400
 #define DUMMY_SEND_MPS_DEFAULT 10000
 
 enum {
@@ -606,7 +611,6 @@ struct user_params_t {
 	int tos;
 	unsigned int lls_usecs;
 	bool lls_is_set;
-	bool dummy_send; //client side only
 	uint32_t dummy_mps; //client side only
 	TicksDuration dummySendCycleDuration; //client side only
 };
--- a/src/SockPerf.cpp
+++ b/src/SockPerf.cpp
@@ -277,10 +277,6 @@ static const AOPT_DESC  common_opt_desc[] =
 		"Set socket accleration before run (available for some of Mellanox systems)"
 	},
 	{
-		OPT_DUMMY_SEND, AOPT_NOARG,	aopt_set_literal( 0 ),	aopt_set_string( "use-dummy" ),
-		"Use dummy send instead of busy wait."
-	},
-	{
 		'd', AOPT_NOARG, aopt_set_literal( 'd' ),	aopt_set_string( "debug" ),
 		"Print extra debug information."
 	},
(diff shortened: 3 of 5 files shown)


src/Switches.h
```diff
@@ -94,7 +95,35 @@ class SwitchOnCycleDuration {
 		}
 		g_cycleStartTime = nextCycleStartTime;
 	}
+
+	inline void execute(int ifd, int buffer_size) {
```
rosenbaumalex: 1. Better to use the original function with default arguments so you can call with and without input values.
2. Also add something line "NOT_IN_USE(x)" marco. Is there something like that in sockperf today?

src/Switches.h
```diff
@@ -94,7 +95,35 @@ class SwitchOnCycleDuration {
...
+		uint8_t buffer[buffer_size];
+		TicksTime nextCycleStartTime = g_cycleStartTime + g_pApp->m_const_params.cycleDuration;
+		while (!g_b_exit) {
+			if (TicksTime::now() >= nextCycleStartTime) {
+				break;
+			}
+
+			sendto(ifd, buffer, sizeof(buffer), MSG_SYN, (struct sockaddr*)& (g_fds_array[ifd]->server_addr), sizeof(struct sockaddr));
```
rosenbaumalex: Currently this dummy send is happening at the max rate allowed which is not  what we expect customers to do in real life. We don't want to send in such a high rate.

Better to have the '--use-dummy' configuration to get a rate, so we can run it with 1000 mps as warmup.
I would recommend having a default rate of 1000mps as --dummy-send-rate, and optionally have user set a different value, like 10kmps

rosenbaumalex: better to create a sockperf level define, something like:
#define DUMMY_SEND_FLAG    MSG_SYN // Used by offload libraries to do egress path warm-up of caches. It is not in use by kernel. WARNING: it will actually end this packet on the wire.

rosenbaumalex: After more thought, a bigger design change...
I think the "send dummy" should be separate from the sleep logic. lets talk about it offline.

rosenbaumalex (changes requested):
After more thought, I want to discuss with you a bigger design change... I think the "send dummy" should be separate from the sleep logic. let's talk about it offline.

src/Switches.h
liranoz12: 1. Yes i agree.
2. No, Sockperf does not have something like "NOT_IN_USE(x)" and i don't sure it is necessary. There are many areas like this in the code :
```
class SwitchOff {
public:
	inline void execute() {}
	inline void execute(int) {}
//	inline void execute(unsigned long long) {}
	inline void execute(uint64_t) {}
	inline void execute(int, uint64_t) {}
	inline void execute(TicksTime & _ticks) {}
	inline void execute(struct sockaddr_in *clt_addr, uint64_t seq_num, bool is_warmup) {}
	inline void execute(Message *pMsgRequest, Message * pMsgReply) {}
	inline void execute(Message *pMsgRequest) {}
};
```

liranoz12:
@rosenbaumalex  Please review

src/Client.h
```diff
@@ -313,7 +313,7 @@ class Client : public ClientBase{
 		m_switchMsgSize.execute(m_pMsgRequest);
 
 		//idle
-		m_switchCycleDuration.execute();
+		m_switchCycleDuration.execute(ifd, m_pMsgRequest->getLength());
```
rosenbaumalex: don't pass lenght, pass the m_pMsg to execute()

src/Defs.h
```diff
@@ -147,6 +147,8 @@ extern const int MAX_FDS_NUM;
 #define DEFAULT_SELECT_TIMEOUT_MSEC	10
 #define DEFAULT_DEBUG_LEVEL			0
 
+#define DUMMY_SEND_FLAG MSG_SYN // Used by offload libraries to do egress path warm-up of caches. It is not in use by kernel. WARNING: it will actually end this packet on the wire.
```
rosenbaumalex: 1.  break comment to NEWLINEs
2. read more in URL.. to github/libvma/... vma_extra.h

src/SockPerf.cpp
```diff
@@ -312,6 +312,10 @@ static const AOPT_DESC  client_opt_desc[] =
 		OPT_OUTPUT_PRECISION, AOPT_NOARG,	aopt_set_literal( 0 ),	aopt_set_string( "increase_output_precision" ),
 		"Increase number of digits after decimal point of the throughput output (from 3 to 9). "
 	},
+	{
+		OPT_DUMMY_SEND, AOPT_OPTARG,	aopt_set_literal( 0 ),	aopt_set_string( "dummy-send" ),
+		"Use dummy-send instead of busy wait. optional: set dummy-send rate per second (default 10000). usage: --dummy-send <rate>"
```
rosenbaumalex: fix:
1. 10,000 (use comma)
2. option: "[<rate>]"
3. add "MAX" as option
4. explain that rate value "must be lower than --mps"

src/SockPerf.cpp
```diff
@@ -2093,6 +2097,26 @@ static int parse_client_opt( const AOPT_OBJECT *client_obj )
 		if ( !rc && aopt_check(client_obj, OPT_OUTPUT_PRECISION) ) {
 			s_user_params.increase_output_precision = true;
 		}
+		if ( !rc && aopt_check(client_obj, OPT_DUMMY_SEND) ) {
+			s_user_params.dummy_send = true;
```
rosenbaumalex: check if can get rid of bool, and use !=0 fro ON and 0 for OFF

src/SockPerf.cpp
```diff
@@ -2290,6 +2314,8 @@ void set_defaults()
 	s_user_params.daemonize = false;
 
 	s_user_params.withsock_accl = false;
+	s_user_params.dummy_send = false;
+	s_user_params.dummy_mps = DUMMY_SEND_MPS_DEFAULT;
```
rosenbaumalex: move set of default to only if command line enabled

src/SockPerf.cpp
```diff
@@ -3205,6 +3231,16 @@ int bringup(const int *p_daemonize)
...
 			rc = SOCKPERF_ERR_BAD_ARGUMENT;
 		}
+
+		if ( !rc && s_user_params.dummy_send && s_user_params.mps == UINT32_MAX) {
+			log_err("Dummy send is not supported when message-rate = MAX");
+			rc = SOCKPERF_ERR_BAD_ARGUMENT;
+		}
+
+		if ( !rc && s_user_params.dummy_send && s_user_params.mps > s_user_params.dummy_mps) {
+			log_err("Dummy send is not supported when message-rate is greater then the dummy-message-rate");
```
rosenbaumalex: error msg: "Dummy send is allowed only if dumm-send-rate is lower than regular msg rate"

src/SockPerf.cpp
```diff
@@ -3205,6 +3231,16 @@ int bringup(const int *p_daemonize)
 			log_msg("Number of threads should be less than sockets count");
 			rc = SOCKPERF_ERR_BAD_ARGUMENT;
 		}
+
+		if ( !rc && s_user_params.dummy_send && s_user_params.mps == UINT32_MAX) {
```
rosenbaumalex: will be removed once bool is removed

src/SockPerf.cpp
```diff
@@ -3205,6 +3231,16 @@ int bringup(const int *p_daemonize)
...
+			log_err("Dummy send is not supported when message-rate = MAX");
+			rc = SOCKPERF_ERR_BAD_ARGUMENT;
+		}
+
+		if ( !rc && s_user_params.dummy_send && s_user_params.mps > s_user_params.dummy_mps) {
```
rosenbaumalex: use >= instead of > so dummy rate must be slightly lower the packet rate ALWAYS

src/Switches.h
```diff
@@ -47,6 +47,7 @@ class SwitchOff {
 	inline void execute(struct sockaddr_in *clt_addr, uint64_t seq_num, bool is_warmup) {}
 	inline void execute(Message *pMsgRequest, Message * pMsgReply) {}
 	inline void execute(Message *pMsgRequest) {}
+	inline void execute(int ifd, int buffer_size) {}
```
rosenbaumalex: instead of buf_size use pMsgRequest

src/Switches.h
```diff
@@ -95,6 +96,33 @@ class SwitchOnCycleDuration {
...
+		uint8_t buffer[buffer_size];
+		TicksTime nextCycleStartTime = g_cycleStartTime + g_pApp->m_const_params.cycleDuration;
+		TicksTime nextDummySendTime = g_cycleStartTime + g_pApp->m_const_params.dummySendCycleDuration;
+		while (!g_b_exit) {
+			if (TicksTime::now() >= nextCycleStartTime) {
+				break;
+			}
+
+			if (TicksTime::now() > nextDummySendTime) {
+				sendto(ifd, buffer, sizeof(buffer), DUMMY_SEND_FLAG, (struct sockaddr*)& (g_fds_array[ifd]->server_addr), sizeof(struct sockaddr));
```
rosenbaumalex: we should consider using a flag which is not supported by OS so this frame gets dropped and does not get queued by OS on TCP and UDP flows.

rosenbaumalex (commented):
1. some small fixes
2. consider using un=-supported flag so send will fail

liranoz12:
@rosenbaumalex Please review

src/Switches.h
```diff
@@ -83,7 +84,7 @@ class SwitchOnMsgSize {
 class SwitchOnCycleDuration {
 public:
 	//busy wait between two cycles starting point and take starting point of next cycle
-	inline void execute() {
+	inline void execute(Message *pMsgRequest, int ifd) {
```
avnerbh: it is better to replace:
  

> execute(Message *pMsgRequest, int ifd)

with:
> execute(Message *, int)

(and if you want, feel free to also fix  above)

src/Switches.h
```diff
@@ -95,6 +96,33 @@ class SwitchOnCycleDuration {
...
 	}
 };
+
+//=============================================================================
+class SwitchOnDummySend {
+public:
+	// dummy send between two cycles starting point and take starting point of next cycle
+
+	inline void execute(Message *pMsgRequest, int ifd) {
+		uint8_t buffer[pMsgRequest->getLength()];
```
avnerbh: *this line is not C code*, but a gcc extension - hence, you must make sure it compiles on all supported platforms, like Windows and BSD 
if there is no additional example in SockPerf for this trick, I recommend that you will use static array of size MAX_MSG_LEN

src/Defs.h
```diff
@@ -147,6 +147,13 @@ extern const int MAX_FDS_NUM;
 #define DEFAULT_SELECT_TIMEOUT_MSEC	10
 #define DEFAULT_DEBUG_LEVEL			0
 
+/*
+Used by offload libraries to do egress path warm-up of caches.
+It is not in use by kernel. WARNING: it will actually end this packet on the wire.
+More info at vma_extra.h
+*/
+#define DUMMY_SEND_FLAG MSG_SYN // equals to 0x400
```
avnerbh, line 155: it can be safer to add a comment - this value is shared with vma (and solarflare); hence, don't change here without changing accordingly in VMA

src/Switches.h
```diff
@@ -95,6 +96,33 @@ class SwitchOnCycleDuration {
...
+		TicksTime nextCycleStartTime = g_cycleStartTime + g_pApp->m_const_params.cycleDuration;
+		TicksTime nextDummySendTime = g_cycleStartTime + g_pApp->m_const_params.dummySendCycleDuration;
+		while (!g_b_exit) {
+			if (TicksTime::now() >= nextCycleStartTime) {
```
avnerbh: keep TT:now() in temp variable instead of calling the function twice

avnerbh (commented):
nice feature, please see my comments

liranoz12:
@rosenbaumalex @avnerbh  Please review

rosenbaumalex (approved):

PR #94, opened 2017-08-02
Modify the log message which indicates that the requested message rate

liranoz12:
is too high.


igor-ivanov:
:+1:

PR #134, opened 2020-08-04
Add write-messages-to-file option.

apegushin:
This extends sockperf with ability to report RTT for a message that is being persisted on the server side with DSYNC and/or DIRECT flags. The purpose of this change is to have sockperf mimic a simple DB server.

### DIFF
--- a/src/server.h
+++ b/src/server.h
@@ -98,6 +98,16 @@ public:
     int server_accept(int ifd);
 
 private:
+    bool b_write_msg_file_init;
+    FILE *write_msg_file;
+    int write_msg_fd;
+    int open_fd_default_flags;
+    uint8_t *write_msg_buf;
+    uint8_t *write_msg_buf_aligned;
+    void setup_write_msg_to_file();
+    void cleanup_write_msg_to_file();
+    void open_fd_align_buffer(int flags);
+
     SwitchActivityInfo m_switchActivityInfo;
     SwitchCalcGaps m_switchCalcGaps;
 };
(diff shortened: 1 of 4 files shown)


src/server.h
```diff
@@ -255,6 +267,38 @@ inline bool Server<IoType, SwitchActivityInfo, SwitchCalcGaps>::server_receive_t
...
+                if (fflush(write_msg_file) != 0) {
+                    exit_with_log("Failed to flush file stream.", SOCKPERF_ERR_FATAL);
+                }
+            } else {
+                assert(s_user_params.write_msg_file_flags_opt != DEFAULT_FILE_FLAGS_OPT);
+
+                memcpy(write_msg_buf_aligned, msg, msg_len);
+                ssize_t wb = write(write_msg_fd, write_msg_buf_aligned, msg_len);
+
+                if (wb != msg_len) {
```
ChrisCoe, line 295: For consistency, cast `msg_len` to `size_t` as done in `if (write_msg_file){ ... }` above.

src/sockperf.cpp
```diff
@@ -1466,6 +1483,57 @@ static int proc_mode_server(int id, int argc, const char **argv) {
...
+                s_user_params.write_msg_file_flags_opt = value;
+
+                /* When not using a file stream to write messages to file, check that Direct IO (O_DIRECT)
+                 * requirements for userspace buffer alignment are met. */
+                if (value != DEFAULT_FILE_FLAGS_OPT) {
+                    if (aopt_check(server_obj, OPT_WRITE_MSG_BUF_ALIGN)) {
+                        const char *optarg = aopt_value(server_obj, OPT_WRITE_MSG_BUF_ALIGN);
+                        long int alignment = strtol(optarg, NULL, 0);
+
+                        if (alignment >= DEFAULT_BUF_ALIGNMENT && ((alignment & (alignment - 1)) == 0)) {
```
ChrisCoe, line 1526: Nice. Maybe add a small comment here that we are checking for powers of 2? Fine either way.

ChrisCoe:
Thanks for adding this write-messages-to-file option. This would be great to have. 

@igor-ivanov can you follow up on this?

PR #157, opened 2021-12-04
Fix null pointer access

dolby360:
In case the value of --mc-source-filter (which should be the IP address) is not found the log message is trying to print the IP address which in this case is definite NULL.

### DIFF
--- a/src/sockperf.cpp
+++ b/src/sockperf.cpp
@@ -1905,7 +1905,7 @@ static int parse_common_opt(const AOPT_OBJECT *common_obj) {
                     }
                 }
             } else {
-                log_msg("'-%d' Invalid address: %s", OPT_MC_SOURCE_IP, optarg);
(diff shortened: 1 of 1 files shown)


ghost:

ChrisCoe:
Nice catch! The current else statement would always access null without your code change. Could the log instead say `Invalid IP address value` or `Invalid value`? This is more consistent with the logs in that file when `if (optarg) { ... } else { ... }` enters else statement.

agalanin-at-nvidia:
Fix for this problem is already included in pull request https://github.com/Mellanox/sockperf/pull/156.

dolby360:
> Fix for this problem is already included in pull request #156.

OK whatever, you can close this PR. That just was my first attempt to get into OS world :)

agalanin-at-nvidia:
Pull request #156 is merged now, so I am closing this PR.

PR #159, opened 2021-12-12
Functions that return SOCKPERF_ERROR are now declared with the right type

dolby360:
Also, compare rc with the correct error code


ghost:

ChrisCoe:
I would actually prefer if return type stays as int, but that's a preference in style. I'd wait to see what the maintainers think. I do agree with the rest of the changes in this file.

PR #161, opened 2021-12-16
issue: 2890836 support IPv6

### LINKED ISSUE #12: issue: 2890836 Support IPv6
```
imported from: http://argus-bg.dnsalias.org/issues/727

What need is this feature going to satisfy?
1. Tool should support IPv6


```

Original issue reported on code.google.com by `igor.ivanov@itseez.com` on 1 Apr 2011 at 2:00

agalanin-at-nvidia:
Summary of changes:
* Added support for IPv4/IPv6/hostname in all places previously supporting only IPv4. In a feedfile IPv6 addresses must be enclosed into square brackets. Resolver prefers IPv6 addresses because :: can listen for both IPv4 and IPv6 clients.
* I use own structure `sockaddr_store_t` to store socket addresses because default sockaddr_storage is 128 bytes long and it is too big for us.
* Improved regexp usage: one regexp per loop instead of one regexp per iteration.

Limitations:
* IPv6 is not supported in VMA SocketXtreme because of API limitations.
* Windows support is broken. Need C++11 (std::regexp) or our sophisticated parser to parse feedfile. Will be fixed in RM#2904933.
* IPv6 multicast support is incomplete. There are no direct equivalents of IPv4 group membership sockopts, need to study more about it. Will be added in RM#2904936.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -860,34 +860,27 @@ int Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration,
     if (rc == SOCKPERF_ERR_NONE) {
         // cycle through all set fds in the array (with wrap around to beginning)
         for (int ifd = m_ioHandler.m_fd_min; ifd <= m_ioHandler.m_fd_max; ifd++) {
+            fds_data *data = g_fds_array[ifd];
 
-            if (!(g_fds_array[ifd] && (g_fds_array[ifd]->active_fd_list))) continue;
+            if (!(data && (data->active_fd_list))) continue;
 
-            struct sockaddr_in *p_client_bind_addr =
-                (struct sockaddr_in *)&g_pApp->m_const_params.client_bind_info;
-            if (p_client_bind_addr->sin_port || p_client_bind_addr->sin_addr.s_addr) {
-                log_dbg("[fd=%d] Binding to: %s:%d...", ifd,
-                        inet_ntoa(p_client_bind_addr->sin_addr),
-                        ntohs(p_client_bind_addr->sin_port));
-                if (bind(ifd, (struct sockaddr *)p_client_bind_addr, sizeof(struct sockaddr)) < 0) {
-                    log_err("[fd=%d] Can`t bind socket %s:%d", ifd,
-                            inet_ntoa(p_client_bind_addr->sin_addr),
-                            ntohs(p_client_bind_addr->sin_port));
+            const sockaddr_store_t *p_client_bind_addr = &g_pApp->m_const_params.client_bind_info;
+            if (p_client_bind_addr->ss_family != AF_UNSPEC) {
+                socklen_t client_bind_addr_len = g_pApp->m_const_params.client_bind_info_len;
+                std::string hostport = sockaddr_to_hostport(p_client_bind_addr);
+                log_dbg("[fd=%d] Binding to: %s...", ifd, hostport.c_str());
+                if (bind(ifd, reinterpret_cast<const sockaddr *>(p_client_bind_addr), client_bind_addr_len) < 0) {
+                    log_err("[fd=%d] Can`t bind socket %s", ifd, hostport.c_str());
                     rc = SOCKPERF_ERR_SOCKET;
                     break;
                 }
-            } else {
-                log_dbg("[fd=%d] Binding to: %s:%d...", ifd,
-                        inet_ntoa(p_client_bind_addr->sin_addr),
-                        ntohs(p_client_bind_addr->sin_port));
             }
-            if (g_fds_array[ifd]->sock_type == SOCK_STREAM) {
-                log_dbg("[fd=%d] Connecting to: %s:%d...", ifd,
-                        inet_ntoa(g_fds_array[ifd]->server_addr.sin_addr),
-                        ntohs(g_fds_array[ifd]->server_addr.sin_port));
+            if (data->sock_type == SOCK_STREAM) {
+                std::string hostport = sockaddr_to_hostport(data->server_addr);
+                log_dbg("[fd=%d] Connecting to: %s...", ifd, hostport.c_str());
 
-                if (connect(ifd, (struct sockaddr *)&(g_fds_array[ifd]->server_addr),
-                            sizeof(struct sockaddr)) < 0) {
+                if (connect(ifd, reinterpret_cast<const sockaddr *>(&(data->server_addr)),
+                            data->server_addr_len) < 0) {
                     if (os_err_in_progress()) {
 #ifdef USING_VMA_EXTRA_API
                         if (g_pApp->m_const_params.fd_handler_type == SOCKETXTREME && g_vma_api) {
@@ -909,8 +902,8 @@ int Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration,
                 }
 #if defined(DEFINED_TLS)
                 if (g_pApp->m_const_params.tls) {
-                    g_fds_array[ifd]->tls_handle = tls_establish(ifd);
-                    if (!g_fds_array[ifd]->tls_handle) {
+                    data->tls_handle = tls_establish(ifd);
+                    if (!data->tls_handle) {
                         rc = SOCKPERF_ERR_SOCKET;
                         break;
                     }
@@ -923,9 +916,10 @@ int Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration,
              * with the rest of the setsockopt
              */
             if (s_user_params.rate_limit > 0 &&
-                sock_set_rate_limit(ifd, s_user_params.rate_limit)) {
+                    sock_set_rate_limit(ifd, s_user_params.rate_limit)) {
+                std::string hostport = sockaddr_to_hostport(data->server_addr);
                 log_err("[fd=%d] failed setting rate limit on address %s\n", ifd,
-                        inet_ntoa(g_fds_array[ifd]->server_addr.sin_addr));
+                        hostport.c_str());
                 rc = SOCKPERF_ERR_SOCKET;
                 break;
             }
--- a/src/common.h
+++ b/src/common.h
@@ -29,21 +29,33 @@
 #ifndef COMMON_H_
 #define COMMON_H_
 
+#include <string>
+
 #include "defs.h"
-//#include "switches.h"
 #include "message.h"
 #include "tls.h"
 
 extern user_params_t s_user_params;
 //------------------------------------------------------------------------------
 void recvfromError(int fd); // take out error code from inline function
+std::string sockaddr_to_hostport(const struct sockaddr *addr);
+inline std::string sockaddr_to_hostport(const struct sockaddr_store_t *addr) {
+    return sockaddr_to_hostport(reinterpret_cast<const sockaddr *>(addr));
+}
+inline std::string sockaddr_to_hostport(const struct sockaddr_in &addr) {
+    return sockaddr_to_hostport(reinterpret_cast<const sockaddr *>(&addr));
+}
+inline std::string sockaddr_to_hostport(const struct sockaddr_store_t &addr) {
+    return sockaddr_to_hostport(reinterpret_cast<const sockaddr *>(&addr));
+}
+bool is_multicast_addr(const sockaddr_store_t &addr);
 void sendtoError(int fd, int nbytes,
-                 const struct sockaddr_in *sendto_addr); // take out error code from inline function
+                 const struct sockaddr *sendto_addr); // take out error code from inline function
 void exit_with_log(int status);
 void exit_with_log(const char *error, int status);
-void exit_with_log(const char *error, int status, fds_data *fds);
+void exit_with_log(const char *error, int status, const fds_data *fds);
 void exit_with_err(const char *error, int status);
-void print_log_dbg(struct in_addr sin_addr, in_port_t sin_port, int ifd);
+void print_log_dbg(struct sockaddr *addr, int ifd);
 
 int set_affinity_list(os_thread_t thread, const char *cpu_list);
 void hexdump(void *ptr, int buflen);
@@ -53,7 +65,7 @@ int read_int_from_sys_file(const char *path);
 // inline functions
 //------------------------------------------------------------------------------
 static inline int msg_sendto(int fd, uint8_t *buf, int nbytes,
-                             const struct sockaddr_in *sendto_addr) {
+                     const struct sockaddr *sendto_addr, socklen_t addrlen) {
     int ret = 0;
     int flags = 0;
 
@@ -93,13 +105,13 @@ static inline int msg_sendto(int fd, uint8_t *buf, int nbytes,
         } else
 #endif /* DEFINED_TLS */
         {
-            ret = sendto(fd, buf, nbytes, flags, (struct sockaddr *)sendto_addr,
-                         sizeof(struct sockaddr));
+            ret = sendto(fd, buf, nbytes, flags, sendto_addr, addrlen);
         }
 
 #if defined(LOG_TRACE_SEND) && (LOG_TRACE_SEND == TRUE)
-        LOG_TRACE("raw", "%s IP: %s:%d [fd=%d ret=%d] %s", __FUNCTION__,
-                  inet_ntoa(sendto_addr->sin_addr), ntohs(sendto_addr->sin_port), fd, ret,
+        std::string hostport = sockaddr_to_hostport(sendto_addr);
+        LOG_TRACE("raw", "%s IP: %s [fd=%d ret=%d] %s", __FUNCTION__,
+                  hostport.c_str(), fd, ret,
                   strerror(errno));
 #endif /* LOG_TRACE_SEND */
 
@@ -145,5 +157,29 @@ static inline int msg_sendto(int fd, uint8_t *buf, int nbytes,
 }
 
 int sock_set_rate_limit(int fd, uint32_t rate_limit);
+
+/** @brief extract port in network byte order from socket address.
+ */
+static inline uint16_t sockaddr_get_portn(const sockaddr_store_t &addr)
+{
+    switch (addr.ss_family) {
+    case AF_INET:
+        return reinterpret_cast<const sockaddr_in &>(addr).sin_port;
+    case AF_INET6:
+        return reinterpret_cast<const sockaddr_in6 &>(addr).sin6_port;
+    }
+    return 0;
+}
+/** @brief set port in network byte order in socket address.
+ */
+static inline void sockaddr_set_portn(sockaddr_store_t &addr, uint16_t port)
+{
+    switch (addr.ss_family) {
+    case AF_INET:
+        reinterpret_cast<sockaddr_in &>(addr).sin_port = port;
+    case AF_INET6:
+        reinterpret_cast<sockaddr_in6 &>(addr).sin6_port = port;
+    }
+}
 
 #endif /* COMMON_H_ */
--- a/src/defs.h
+++ b/src/defs.h
@@ -38,8 +38,6 @@
 #include <Winbase.h>
 #include <stdint.h>
 
-typedef uint16_t in_port_t;
-
 #else
 
 #ifdef __linux__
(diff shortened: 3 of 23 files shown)


igor-ivanov:
bot:retest

src/client.cpp
```diff
@@ -860,34 +860,27 @@ int Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration,
...
+                if (bind(ifd, (struct sockaddr *)p_client_bind_addr, client_bind_addr_len) < 0) {
+                    log_err("[fd=%d] Can`t bind socket %s", ifd, hostport.c_str());
                     rc = SOCKPERF_ERR_SOCKET;
                     break;
                 }
-            } else {
-                log_dbg("[fd=%d] Binding to: %s:%d...", ifd,
-                        inet_ntoa(p_client_bind_addr->sin_addr),
-                        ntohs(p_client_bind_addr->sin_port));
             }
```
igor-ivanov, line 877: Pay attention that probably logic for 0.0.0.0/0 might be different after your change

src/client.cpp
```diff
@@ -860,34 +860,27 @@ int Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration,
...
-                        inet_ntoa(g_fds_array[ifd]->server_addr.sin_addr),
-                        ntohs(g_fds_array[ifd]->server_addr.sin_port));
+            if (data->sock_type == SOCK_STREAM) {
+                std::string hostport = sockaddr_to_hostport(data->server_addr);
+                log_dbg("[fd=%d] Connecting to: %s...", ifd, hostport.c_str());
 
-                if (connect(ifd, (struct sockaddr *)&(g_fds_array[ifd]->server_addr),
-                            sizeof(struct sockaddr)) < 0) {
+                if (connect(ifd, (struct sockaddr *)&(data->server_addr),
+                            data->server_addr_len) < 0) {
```
igor-ivanov, line 883: just check if `server_addr_len` assigned in all place when `g_fds_array` is initialized?

src/defs.h
```diff
@@ -486,21 +476,31 @@ struct SocketRecvData {
...
 };
 
+// big enough to store sockaddr_in and sockaddr_in6
+struct sockaddr_storage_sp {
+    union {
+        sa_family_t ss_family;
+        sockaddr_in addr4;
+        sockaddr_in6 addr6;
+    };
+};
```
igor-ivanov, line 486: please consider renaming `sockaddr_storage_sp` with `sockaddr_store_t` to meet de facto existing style. For example `struct user_params_t, work_mode_t, os_cpuset_t` etc

src/ip_address.h
```diff
@@ -0,0 +1,139 @@
...
+        switch (key1.m_family) {
+        case AF_INET: {
+            return key1.m_addr4.s_addr == key2.m_addr4.s_addr;
+        }
+        case AF_INET6: {
+            return IN6_ARE_ADDR_EQUAL(&key1.m_addr6, &key2.m_addr6);
+        }
+        }
+        return true;
+    }
```
igor-ivanov, line 121: please consider more compact:
```
    if (key1.m_family == key2.m_family) {
        switch (key1.m_family) {
        case AF_INET:
            return key1.m_addr4.s_addr == key2.m_addr4.s_addr;
        case AF_INET6:
            return IN6_ARE_ADDR_EQUAL(&key1.m_addr6, &key2.m_addr6);
        }
    }
    return false;
```

src/ip_address.h
```diff
@@ -0,0 +1,139 @@
...
+// hash/equal_to functions, by ourself.
+namespace std {
+#if !defined(WIN32) && !defined(__FreeBSD__)
+namespace tr1 {
+#endif
+template <> struct hash<IPAddress> : public std::unary_function<IPAddress, int> {
+    int operator()(IPAddress const &key) const
+    {
+        return key.addr4().s_addr & 0xFF;
+    }
```
igor-ivanov, line 144: it is not clear

src/server.cpp
```diff
@@ -251,7 +259,7 @@ int Server<IoType, SwitchActivityInfo, SwitchCalcGaps>::server_accept(int ifd) {
             log_err("Failed to allocate memory with malloc()");
             return (int)INVALID_SOCKET; // TODO: use SOCKET all over the way and avoid this cast
         }
-        memcpy(tmp, g_fds_array[ifd], sizeof(struct fds_data));
+        tmp = g_fds_array[ifd];
```
igor-ivanov: ??? it seems not equal

src/server.h
```diff
@@ -273,19 +285,29 @@ inline bool Server<IoType, SwitchActivityInfo, SwitchCalcGaps>::handle_message(i
...
             /* always send to the same port recved from */
-            sendto_addr.sin_port = recvfrom_addr.sin_port;
+            switch (recvfrom_addr.ss_family) {
+            case AF_INET:
+                ((sockaddr_in &)sendto_addr).sin_port = ((sockaddr_in &)recvfrom_addr).sin_port;
+                break;
+            case AF_INET6:
+                ((sockaddr_in6 &)sendto_addr).sin6_port = ((sockaddr_in6 &)recvfrom_addr).sin6_port;
+                break;
+            }
```
igor-ivanov: Please consider introducing 'get/set port' functions. Pay attention that get and set value in netorder.

src/sockperf.cpp
```diff
@@ -1723,12 +1684,62 @@ static int proc_mode_server(int id, int argc, const char **argv) {
...
+        bool is_server_mode, sockaddr *addr, socklen_t &addr_len)
+{
+    struct addrinfo hints;
+
+    memset(&hints, 0, sizeof(hints));
+    // allow IPv4 or IPv6
+    hints.ai_family = AF_UNSPEC;
+    // return addresses only from configured address families
+    hints.ai_flags = AI_ADDRCONFIG;
+    // if host is NULL then server address will be 0.0.0.0/:: and client address will be 127.0.0.1/::1
```
igor-ivanov, line 1708: in current flow host can not be NULL.
and instead of introducing `is_server_mode ` host` variable can be used.
```
if (host == NULL) hints.ai_flags |= AI_PASSIVE;
```

I prefer not to change `parse_common_opt` prototype

igor-ivanov (commented):
can we return error with message if ipv6 multicast ip is detected in user passed configuration (command line and fedd file)?

src/sockperf.cpp
agalanin-at-nvidia: `host_str` in `parse_common_opt` can be `NULL` if `-i` option is not specified. This is rare but valid case.

src/ip_address.h
agalanin-at-nvidia: In my code two AF_UNSPEC addresses are equal, in your code they are different.

src/client.cpp
agalanin-at-nvidia: I use AF_UNSPEC family instead of 0.0.0.0:0 as a special value for unset address.

src/server.cpp
agalanin-at-nvidia: oops, didn't notice that they were pointers.

src/ip_address.h
agalanin-at-nvidia: I added hashcode for ipv6 too

src/client.cpp
agalanin-at-nvidia: I think that all cases are covered.

But your comment got me thinking about adding address length validation into IPAddress constructor. I added it.

src/defs.h
igor-ivanov: resolved?

src/ip_address.h
igor-ivanov: not understand

src/defs.h
agalanin-at-nvidia: Renamed to `sockaddr_store_t`.

src/ip_address.h
agalanin-at-nvidia: For TCP connections recvfrom address length is 0, so I create IPAddress with family AF_UNSPEC. All of AF_UNSPEC addresses must be equal, so my variant of comparator returns `true` if key1.family == key2.family == AF_UNSPEC.

agalanin-at-nvidia: rewritten to highlight AF_UNSPEC case

src/sockperf.cpp
agalanin-at-nvidia: rewritten without this argument

igor-ivanov:
bot:retest

igor-ivanov (approved):
commits cleanup is needed

agalanin-at-nvidia:
bot:retest

agalanin-at-nvidia:
bot:retest

agalanin-at-nvidia:
bot:retest

src/common.h
```diff
@@ -146,4 +158,28 @@ static inline int msg_sendto(int fd, uint8_t *buf, int nbytes,
...
+    case AF_INET:
+        return ((sockaddr_in &)addr).sin_port;
+    case AF_INET6:
+        return ((sockaddr_in6 &)addr).sin6_port;
+    }
+    return 0;
+}
+/** @brief set port in network byte order in socket address.
+ */
+static inline void sockaddr_set_portn(const sockaddr_store_t &addr, uint16_t port)
```
AlexanderGrissik: The addr parameter should be a non-const

src/common.h
```diff
@@ -146,4 +158,28 @@ static inline int msg_sendto(int fd, uint8_t *buf, int nbytes,
 
 int sock_set_rate_limit(int fd, uint32_t rate_limit);
 
+/** @brief extract port in network byte order from socket address.
+ */
+static inline uint16_t sockaddr_get_portn(const sockaddr_store_t &addr)
+{
+    switch (addr.ss_family) {
+    case AF_INET:
+        return ((sockaddr_in &)addr).sin_port;
```
AlexanderGrissik: Cast to const sockaddr_in & ...

src/defs.h
```diff
@@ -536,53 +536,43 @@ namespace std {
...
-template <>
-struct hash<struct port_and_type> : public std::unary_function<struct port_and_type, int> {
-    int operator()(struct port_and_type const &key) const {
+template <> struct hash<struct sockaddr_store_t> : public std::unary_function<struct sockaddr_store_t, int> {
+    int operator()(struct sockaddr_store_t const &key) const {
         // XOR "a.b" part of "a.b.c.d" address with 16bit port; leave "c.d" part untouched for
         // maximum hashing
-        return key.sock_type ^ key.port;
+        // use only IPv4 part for hashing
+        return ((sockaddr_in &)key).sin_addr.s_addr ^ ((sockaddr_in &)key).sin_port;
```
AlexanderGrissik: Would this be enough for IPv6?

src/defs.h
```diff
@@ -536,53 +536,43 @@ namespace std {
...
+            const sockaddr_in &k1 = (const sockaddr_in &)key1;
+            const sockaddr_in &k2 = (const sockaddr_in &)key2;
+            return k1.sin_port == k2.sin_port &&
+                k1.sin_addr.s_addr == k2.sin_addr.s_addr;
+        }
+        case AF_INET6: {
+            const sockaddr_in6 &k1 = (const sockaddr_in6 &)key1;
+            const sockaddr_in6 &k2 = (const sockaddr_in6 &)key2;
+            return k1.sin6_port == k2.sin6_port &&
+                IN6_ARE_ADDR_EQUAL(&k1.sin6_addr, &k2.sin6_addr);
```
AlexanderGrissik, line 580: Many impls do this with four 32bit comparisions. We can do it with two 64bit comparisions.

src/iohandlers.cpp
```diff
@@ -28,6 +28,30 @@
 
 #include "iohandlers.h"
 
+static void print_addresses(const fds_data *data, int &list_count)
+{
+    {
+        char hbuf[NI_MAXHOST] = "(unknown)";
+        char pbuf[NI_MAXSERV] = "(unk)";
+        getnameinfo((sockaddr *)&data->server_addr, data->server_addr_len,
```
AlexanderGrissik: These constructs repeat themselves many time. Can we arrange them into a method returning std:string? It is only for logging anyway.

src/ip_address.cpp
```diff
@@ -0,0 +1,103 @@
...
+    int res = getaddrinfo(str, NULL, &hints, &result);
+    if (res == 0) {
+        out.m_family = AF_UNSPEC;
+        for (addrinfo *rp = result; rp != NULL; rp = rp->ai_next) {
+            if (rp->ai_family == AF_INET6) {
+                out.m_family = rp->ai_family;
+                out.m_addr6 = ((sockaddr_in6 *)rp->ai_addr)->sin6_addr;
+            }
+        }
+        if (out.m_family == AF_UNSPEC) {
```
AlexanderGrissik, line 59: Is this to give IPv6  priority?

src/ip_address.cpp
```diff
@@ -0,0 +1,103 @@
...
+        err = gai_strerror(res);
+    }
+    freeaddrinfo(result);
+
+    return res == 0;
+}
+
+IPAddress IPAddress::zero()
+{
+    IPAddress res;
```
AlexanderGrissik, line 75: Not sure if its used in DP, but it can be a static and returned as const IPAddress&

src/ip_address.h
```diff
@@ -0,0 +1,158 @@
...
+    {
+    }
+
+    IPAddress(const IPAddress &rhs) : m_family(rhs.m_family), m_addr6(rhs.m_addr6)
+    {
+    }
+
+    IPAddress(const sockaddr *sa, socklen_t len)
+    {
+        if (len < sizeof(m_family)) {
```
AlexanderGrissik, line 66: m_family is garbage at this stage

src/ip_address.h
```diff
@@ -0,0 +1,158 @@
...
+namespace std {
+#if !defined(WIN32) && !defined(__FreeBSD__)
+namespace tr1 {
+#endif
+template <> struct hash<IPAddress> : public std::unary_function<IPAddress, int> {
+    int operator()(IPAddress const &key) const
+    {
+        switch (key.family()) {
+        case AF_INET:
+            return key.addr4().s_addr & 0xFF;
```
AlexanderGrissik, line 136: Why do we use only 1 byte?

src/ip_address.h
```diff
@@ -0,0 +1,158 @@
...
+        case AF_INET6:
+            return (reinterpret_cast<const uint32_t *>(&key.addr6()))[3];
```
AlexanderGrissik: Would this be enough for IPv6?

src/ip_address.h
```diff
@@ -0,0 +1,158 @@
...
+    {
+        if (len < sizeof(m_family)) {
+            m_family = AF_UNSPEC;
+            return;
+        }
+        m_family = sa->sa_family;
+        switch (m_family) {
+        case AF_INET:
+            if (len >= sizeof(sockaddr_in)) {
+                m_addr4 = ((sockaddr_in *)sa)->sin_addr;
```
AlexanderGrissik: Why not using C++ casts for new code?

src/defs.h
```diff
@@ -486,21 +476,31 @@ struct SocketRecvData {
...
     int is_multicast;               /**< if this socket is multicast */
     int sock_type;                  /**< SOCK_STREAM (tcp), SOCK_DGRAM (udp), SOCK_RAW (ip) */
     int next_fd;
     int active_fd_count; /**< number of active connections (by default 1-for UDP; 0-for TCP) */
     int *
     active_fd_list; /**< list of fd related active connections (UDP has the same fd by default) */
-    struct sockaddr_in *memberships_addr; /**< more servers on the same socket information */
-    struct in_addr mc_source_ip_addr;     /**< message source ip for multicast packet filtering */
+    struct sockaddr_store_t *memberships_addr; /**< more servers on the same socket information */
+    IPAddress mc_source_ip_addr;    /**< message source ip for multicast packet filtering */
```
AlexanderGrissik, line 503: This addition causes a compilation warning with 'memset(tmp, 0, sizeof(struct fds_data));' sockperf.cpp:3046,3346
warning: ‘void* memset(void*, int, size_t)’ clearing an object of non-trivial type ‘fds_data’ {aka ‘struct fds_data’}; use assignment or value-initialization instead [-Wclass-memaccess]
Cast to void* of fs_data* should help.

src/ip_address.h
agalanin-at-nvidia: We are using this hash function for years. I just moved it into another place of code.

src/ip_address.cpp
agalanin-at-nvidia: yes

src/defs.h
agalanin-at-nvidia: Are you sure that data is always 8-byte aligned?

src/defs.h
agalanin-at-nvidia: Added cast to `void *`. In a future (RM 2904933) this will be replaced with default initializers for all struct members.

src/ip_address.cpp
agalanin-at-nvidia: It's only used in command-line parsing code in assignment. So using a static member will not provide any performance improvement.

src/defs.h
AlexanderGrissik: I mean, we can compare the sin6_addr using two 64 bits ==. Alignment should not affect the sin6_addr itself.

The usual impl:
define IN6_ARE_ADDR_EQUAL(a,b) \
  (__extension__							      \
   ({ const struct in6_addr *__a = (const struct in6_addr *) (a);	      \
      const struct in6_addr *__b = (const struct in6_addr *) (b);	      \
      __a->__in6_u.__u6_addr32[0] == __b->__in6_u.__u6_addr32[0]	      \
      && __a->__in6_u.__u6_addr32[1] == __b->__in6_u.__u6_addr32[1]	      \
      && __a->__in6_u.__u6_addr32[2] == __b->__in6_u.__u6_addr32[2]	      \
      && __a->__in6_u.__u6_addr32[3] == __b->__in6_u.__u6_addr32[3]; }))

What could be:
      __a->__in6_u.__u6_addr64[0] == __b->__in6_u.__u6_addr64[0]	      \
      && __a->__in6_u.__u6_addr64[1] == __b->__in6_u.__u6_addr64[1];

agalanin-at-nvidia: I can't imagine how to express it in code without casting a casting a sin6_addr pointer to uint64_t pointer.

src/iohandlers.cpp
agalanin-at-nvidia: I replaced most of these to calls to `sokaddr_to_hostport`. Now instead of
```
IP = 127.0.0.1 PORT = 11111 # TCP sockperf: ERROR: A connection was forcibly closed by a peer (errno=0 Success)
IP = ::1 PORT =11111 # TCP sockperf: ERROR: A connection was forcibly closed by a peer (errno=0 Success)
```
user will see
```
ADDR = 127.0.0.1:11111 # TCP sockperf: ERROR: A connection was forcibly closed by a peer (errno=0 Success)
ADDR = [::1]:11111 # TCP sockperf: ERROR: A connection was forcibly closed by a peer (errno=0 Success)
```
I intentionally left `print_address()` unchanged because this function prints fancy formatted columns:
```
[ 0] IP = 1.1.64.30       PORT = 12345 # UDP
[ 1] IP = 1.1.64.30       PORT = 12346 # UDP
```

src/defs.h
agalanin-at-nvidia: Added other hash for IPv6. Made IPv4 hashing correspond to comment.

src/ip_address.h
agalanin-at-nvidia: updated

src/defs.h
AlexanderGrissik: I believe we can cast..

AlexanderGrissik (approved):

src/defs.h
agalanin-at-nvidia: Imagine that &addr == 0x100004. This is an aligned address for 32-bit integer `__in6_u.__u6_addr32[0]` but for 64-bit integer `__in6_u.__u6_addr64[0]` this address is not aligned. This will result in SIGBUS on SPARC and slower access on another architectures.

igor-ivanov:
bot:retest

agalanin-at-nvidia:
bot:retest

PR #165, opened 2022-02-01
issue: 1308155 Enhance full-log output

EldarShalev:
fixing filter.awk tool
fixing commit 1a20febc4e3cb2ff30ae362a0ae2f86116550f00
The mentioned commit added to the output the packet number and the latency, but caused a bug to filter.awk
Fixing filter.awk to support new representation and the parameters to be in order. 

Signed-off-by: Eldar Shalev <eldarsh@nvidia.com>

### DIFF
--- a/tools/filter.awk
+++ b/tools/filter.awk
@@ -7,13 +7,13 @@ BEGIN {
 	dataStarted = 0
 }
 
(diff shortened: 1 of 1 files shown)


EldarShalev:
>

igor-ivanov (approved):

PR #171, opened 2022-03-14
issue: 3002911 Fix error when compiling OpenSSL 3.0.0

alexbriskin:
OpenSSL development(alpha) versions for 3.0.0 release deprecated lot's
of previously supported APIs before adding the new ones.
Additionally, there is no simple way to distinguish between 3.0.0 alpha
and release versions since the OPENSSL_VERSION_NUMBER is of the same
value. Thus, the only reasonable solution seems to be the one where we
migrate to the OpenSSL 3 APIs for versions that after the 3.0.0.
Potential "API deprecated" warnings will be suppressed for previous
versions.

Signed-off-by: Alex Briskin <abriskin@nvidia.com>

### DIFF
--- a/src/tls.cpp
+++ b/src/tls.cpp
@@ -121,7 +121,7 @@ int tls_init(void) {
         goto error_free_ctx;
     }
 
-#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+#if OPENSSL_VERSION_NUMBER > 0x30000000L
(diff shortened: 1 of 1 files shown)


alexbriskin:
PR withdrawn. Socperf users should either use a stable OpenSSL 3.0.0 or OpenSSL 1.1.1

PR #189, opened 2022-06-24
issue: 3115377 Add copyright headers check

vasily-v-ryabov:
Signed-off-by: Vasily Ryabov <vryabov@nvidia.com>


.ci/proj_jjb.yaml
```diff
@@ -48,15 +48,15 @@
...
             description: "Run static analysis using csbuild tool."
         - bool:
             name: "do_style"
-            default: true
+            default: false
             description: "Analysis source code for coding style."
         - bool:
             name: "do_coverity"
-            default: true
+            default: false
```
agalanin-at-nvidia, line 59: Why do you disable these checks by default? May be you forgot to revert debugging changes.

contrib/test_jenkins.sh
```diff
@@ -213,6 +214,16 @@ for target_v in "${target_list[@]}"; do
 	        rc=$((rc + $ret))
 	    fi
     fi
+    if [ 9 -lt "$jenkins_opt_exit" -o "$rc" -eq 0 ]; then
```
agalanin-at-nvidia, line 217: I think that it is better to move this quick check before compilation and unit tests to speed-up error detection.

contrib/jenkins_tests/copyrights.sh
```diff
@@ -0,0 +1,51 @@
...
+  --config ${WORKSPACE}/contrib/jenkins_tests/copyright-check-map.yaml \
+  --path ${WORKSPACE} \
+  --git-repo ${WORKSPACE} | tee copyrights.log
+exit_code=$?
+echo "exit_code=${exit_code}"
+# Correct error code is not returned by the script
+set +eE
+grep -rn ERROR copyrights.log
+exit_code=$?
+set -eE
```
agalanin-at-nvidia, line 36: If tool is failed to start then error log can be empty. So I added check for first exit code too. BTW, have you tried `set -o pipefail` to check exit codes for all commands in pipeline?
```suggestion
exit_code=$?
echo "exit_code=${exit_code}"
if [ $exit_code -eq 0 ]
then
    # Script does not return error flag in exit code
    set +eE
    grep ERROR copyrights.log
    exit_code=$?
    set -eE
fi
```

agalanin-at-nvidia (approved):

vasily-v-ryabov:
@igor-ivanov I wanted to fix some Alexander's comments (especially the first one). OK, I can do it in next PR.

PR #193, opened 2022-07-20
issue: 2935075 Fix Windows support

EldarShalev:
Adding Windows support for sockperf including AF_UNIX. 
Most of the code is added with #ifdef WIN32, so no impact is expected on linux.

Major changes include:

1. AF_UNIX which in windows is only supported by SOCK_STREAM mode so no binding is needed from client side (thus no path_pid_num binding as provided in linux) 
2. Using new API s.a 'WSAStartup', 'DeleteFileA' (instead of unlink), 'closesocket()' instead of close(), etc.. 
3. Working with <winsock2.h> 

More about AF_UNIX in windows: 
https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/

### DIFF
--- a/doc/main.dox
+++ b/doc/main.dox
@@ -110,7 +110,9 @@ sockperf: [tid 4805] using epoll() to block on socket(s)
    - address - Internet Protocol (IP) address or host name (IPv6 addresses must be enclosed in square brackets);
    - port - Port number;
    - mc_src_addr - Optional multicast source IP address or host name.
-   - PATH - absolute path for UNIX Domain Socket, the line must start with '/'
+   - PATH - absolute path for UNIX Domain Socket
+            Linux - line must start with '/'
+            Windows - line must start with absoulte path including directory, e.g 'c:\tmp\test'
 
 
 @subsection _option 3.1 Available options
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -204,7 +204,7 @@ void printHistogram(uint32_t binSize, std::map<uint32_t, uint32_t> &activeBins, 
     std::map<uint32_t, uint32_t>::iterator itr;
 
     // Scale to terminal
-#ifndef WIN32
+#ifndef __windows__
     struct winsize size;
     ioctl(STDOUT_FILENO, TIOCGWINSZ, &size);
     terminalWidth = size.ws_col;
@@ -824,7 +824,7 @@ static int _connect_check(int ifd) {
 
 #define POLL_TIMEOUT_MS 1
 
-#ifdef WIN32
+#ifdef __windows__
     fd_set rfds, wfds;
     struct timeval tv;
     tv.tv_sec = 0;
@@ -841,7 +841,7 @@ static int _connect_check(int ifd) {
     struct pollfd fds = { .fd = ifd, .events = POLLIN | POLLOUT, };
     pollrc = poll(&fds, 1, POLL_TIMEOUT_MS);
     avail = pollrc > 0 && (fds.revents & (POLLIN | POLLOUT));
-#endif /* WIN32 */
+#endif /* __windows__ */
 
     if (pollrc < 0) {
         log_err("Failed to poll for events during connection establishment");
@@ -884,11 +884,11 @@ int Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration,
                 struct sockaddr_store_t unix_addr;
                 socklen_t unix_addr_len;
                 if (p_client_bind_addr->ss_family == AF_UNIX && g_pApp->m_const_params.sock_type == SOCK_DGRAM) { // Need to bind localy
-#ifdef WIN32
+#ifdef __windows__
                     log_err("AF_UNIX with DGRAM isn't supported in windows");
                     rc = SOCKPERF_ERR_SOCKET;
                     break;
-#else
+#else // __windows__
                     if (p_client_bind_addr->addr_un.sun_path[0] == 0) { // no specific addr client_info was provoided
                         std::string sun_path = build_client_socket_name(&s_user_params.addr.addr_un, getpid(), ifd);
                         log_dbg("No client name was provided, setting addr_un.sun_path to %s\n",
@@ -906,7 +906,7 @@ int Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration,
                         p_client_bind_addr = &unix_addr;
                         client_bind_addr_len = unix_addr_len;
                     }
-#endif
+#endif //__windows__
                 }
 
                 log_dbg("[fd=%d] Binding to: %s...", ifd, hostport.c_str());
@@ -1258,7 +1258,7 @@ void client_handler(handler_info *p_info) {
             client_handler<IoRecvfromMUX>(p_info->fd_min, p_info->fd_max, p_info->fd_num);
             break;
         }
-#ifndef WIN32
+#ifndef __windows__
         case POLL: {
             client_handler<IoPoll>(p_info->fd_min, p_info->fd_max, p_info->fd_num);
             break;
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -211,7 +211,7 @@ void printHistogram(uint32_t binSize, std::map<uint32_t, uint32_t> &activeBins, 
 #else
     CONSOLE_SCREEN_BUFFER_INFO csbi;
     GetConsoleScreenBufferInfo(GetStdHandle(STD_OUTPUT_HANDLE),&csbi);
-    terminal_width = csbi.dwSize.X;
+    terminalWidth = csbi.dwSize.X;
 #endif
     for(itr = activeBins.begin(); itr != activeBins.end(); ++itr) {
         currFrequency = itr->second;
@@ -826,7 +826,9 @@ static int _connect_check(int ifd) {
 
 #ifdef WIN32
     fd_set rfds, wfds;
-    struct timeval tv = { .tv_sec = 0, .tv_usec = POLL_TIMEOUT_MS * 1000, };
+    struct timeval tv;
+    tv.tv_sec = 0;
+    tv.tv_usec = POLL_TIMEOUT_MS * 1000;
 
     FD_ZERO(&rfds);
     FD_ZERO(&wfds);
@@ -882,6 +884,11 @@ int Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration,
                 struct sockaddr_store_t unix_addr;
                 socklen_t unix_addr_len;
                 if (p_client_bind_addr->ss_family == AF_UNIX && g_pApp->m_const_params.sock_type == SOCK_DGRAM) { // Need to bind localy
+#ifdef WIN32
+                    log_err("AF_UNIX with DGRAM isn't supported in windows");
+                    rc = SOCKPERF_ERR_SOCKET;
+                    break;
+#else
                     if (p_client_bind_addr->addr_un.sun_path[0] == 0) { // no specific addr client_info was provoided
                         std::string sun_path = build_client_socket_name(&s_user_params.addr.addr_un, getpid(), ifd);
                         log_dbg("No client name was provided, setting addr_un.sun_path to %s\n",
@@ -899,7 +906,9 @@ int Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration,
                         p_client_bind_addr = &unix_addr;
                         client_bind_addr_len = unix_addr_len;
                     }
+#endif
                 }
+
                 log_dbg("[fd=%d] Binding to: %s...", ifd, hostport.c_str());
                 if (bind(ifd, reinterpret_cast<const sockaddr *>(p_client_bind_addr), client_bind_addr_len) < 0) {
                     log_err("[fd=%d] Can`t bind socket %s", ifd, hostport.c_str());
@@ -1078,9 +1087,7 @@ void Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration
                 // Disarm timer
                 timer.it_value.tv_sec = 0;
                 timer.it_value.tv_usec = 0;
-                if (setitimer(ITIMER_REAL, &timer, NULL)) {
-                    log_err("ERROR: setitimer() failed when disarming");
-                }
+                os_set_disarm_timer(timer);
                 log_msg("Test end (reached packet number)");
                 g_b_exit = true;
             }
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -64,7 +64,7 @@ std::string sockaddr_to_hostport(const struct sockaddr *addr)
     if (addr->sa_family == AF_INET6) {
         return "[" + std::string(hbuf) + "]:" + std::string(pbuf);
     } else if (addr->sa_family == AF_UNIX) {
-        return std::string(pbuf) + " [UNIX]";
+        return std::string(addr->sa_data);
     } else {
         return std::string(hbuf) + ":" + std::string(pbuf);
     }
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -231,7 +231,7 @@ void hexdump(void *ptr, int buflen) {
 const char *handler2str(fd_block_handler_t type) {
     // must be coordinated with fd_block_handler_t in defs.h
     static const char *s_fds_handle_desc[FD_HANDLE_MAX] = { "recvfrom", "recvfrom", "select"
-#ifndef WIN32
+#ifndef __windows__
                                                             ,
                                                             "poll", "epoll",
 #ifdef USING_VMA_EXTRA_API
--- a/src/common.h
+++ b/src/common.h
@@ -82,7 +82,7 @@ static inline int msg_sendto(int fd, uint8_t *buf, int nbytes,
  * the MSG_NOSIGNAL flag.
  * Note: another way is call signal (SIGPIPE,SIG_IGN);
  */
-#ifndef WIN32
+#ifndef __windows__
     flags = MSG_NOSIGNAL;
 
     /*
@@ -94,7 +94,7 @@ static inline int msg_sendto(int fd, uint8_t *buf, int nbytes,
     if (g_pApp->m_const_params.is_nonblocked_send) {
         flags |= MSG_DONTWAIT;
     }
-#endif
+#endif // __windows__
 
     int size = nbytes;
     if (g_fds_array[fd]->sock_type == SOCK_STREAM) {
--- a/src/defs.h
+++ b/src/defs.h
@@ -31,7 +31,7 @@
 
 #define __STDC_FORMAT_MACROS
 
-#ifdef WIN32
+#ifdef __windows__
 #include <WS2tcpip.h>
 #include <Winsock2.h>
 #include <unordered_map>
@@ -96,7 +96,7 @@ typedef unsigned short int sa_family_t;
 #include "playback.h"
 #include "ip_address.h"
 
-#if !defined(WIN32) && !defined(__FreeBSD__)
+#if !defined(__windows__) && !defined(__FreeBSD__)
 #include "vma-xlio-redirect.h"
 #ifdef USING_VMA_EXTRA_API
 #include <mellanox/vma_extra.h>
@@ -152,7 +152,7 @@ const uint32_t TEST_FIRST_CONNECTION_FIRST_PACKET_TTL_THRESHOLD_MSEC = 50;
     ":(6553[0-5]|655[0-2][0-9]|65[0-4][0-9]{2}|6[0-4][0-9]{3}|[0-5]?[0-9]{1,4})"                   \
     "(:([a-zA-Z0-9\\.\\-]+|\\[([a-fA-F0-9.:]+)\\]))?[\r\n]"
 
-#ifdef WIN32
+#ifdef __windows__
 #define UNIX_DOMAIN_SOCKET_FORMAT_REG_EXP                                                          \
         "^[UuTt]:([A-Za-z]:[\\\\/].*)[\r\n]*"
 #define RESOLVE_ADDR_FORMAT_SOCKET                                                                 \
@@ -635,7 +635,7 @@ typedef enum { // must be coordinated with s_fds_handle_desc in common.cpp
     RECVFROM = 0,
     RECVFROMMUX,
     SELECT,
-#ifndef WIN32
+#ifndef __windows__
     POLL,
     EPOLL,
 #endif
--- a/src/defs.h
+++ b/src/defs.h
@@ -33,10 +33,12 @@
 
 #ifdef WIN32
 #include <WS2tcpip.h>
-#include <sys/types.h>
+#include <Winsock2.h>
 #include <unordered_map>
 #include <Winbase.h>
 #include <stdint.h>
+#include <afunix.h>
+typedef unsigned short int sa_family_t;
 
 #else
 
@@ -60,7 +62,6 @@
 #include <stdint.h>
 #include <unordered_map>
 #include <unistd.h>   /* getopt() and sleep()*/
-#include <inttypes.h> /* printf PRItn */
 #include <poll.h>
 #include <pthread.h>
 #include <sys/time.h>   /* timers*/
@@ -75,6 +76,7 @@
 
 #endif
 
+#include <inttypes.h> /* printf PRItn */
 #include <stdlib.h> /* random()*/
 #include <string.h>
 #include <stdio.h>
@@ -150,8 +152,17 @@ const uint32_t TEST_FIRST_CONNECTION_FIRST_PACKET_TTL_THRESHOLD_MSEC = 50;
     ":(6553[0-5]|655[0-2][0-9]|65[0-4][0-9]{2}|6[0-4][0-9]{3}|[0-5]?[0-9]{1,4})"                   \
     "(:([a-zA-Z0-9\\.\\-]+|\\[([a-fA-F0-9.:]+)\\]))?[\r\n]"
 
+#ifdef WIN32
 #define UNIX_DOMAIN_SOCKET_FORMAT_REG_EXP                                                          \
-         "^[UuTt]:(/.+)[\r\n]"
+        "^[UuTt]:([A-Za-z]:[\\\\/].*)[\r\n]*"
+#define RESOLVE_ADDR_FORMAT_SOCKET                                                                 \
+        "[A-Za-z]:[\\\\/].*"
+#elif __linux__
+#define UNIX_DOMAIN_SOCKET_FORMAT_REG_EXP                                                          \
+        "^[UuTt]:(/.+)[\r\n]"
+#define RESOLVE_ADDR_FORMAT_SOCKET                                                                 \
+        "/.+"
+#endif
 
 #define PRINT_PROTOCOL(type)                                                                       \
     ((type) == SOCK_DGRAM ? "UDP" : ((type) == SOCK_STREAM ? "TCP" : "<?>"))
@@ -627,8 +638,8 @@ typedef enum { // must be coordinated with s_fds_handle_desc in common.cpp
 #ifndef WIN32
     POLL,
     EPOLL,
-    SOCKETXTREME,
 #endif
+    SOCKETXTREME,
     FD_HANDLE_MAX } fd_block_handler_t;
 
 struct user_params_t {
--- a/src/ip_address.cpp
+++ b/src/ip_address.cpp
@@ -29,9 +29,13 @@
 #include "ip_address.h"
 #include "common.h"
 
+#ifdef WIN32
+#include <ws2tcpip.h>
+#else
 #include <arpa/inet.h>
 #include <netdb.h>
 #include <string.h>
+#endif
 
 IPAddress::IPAddress(const IPAddress &rhs) : m_family(rhs.m_family), m_addr6(rhs.m_addr6), m_addr_un(rhs.m_addr_un)
 {
@@ -100,7 +104,7 @@ bool IPAddress::resolve(const char *str, IPAddress &out, std::string &err)
             }
         }
     } else {
-        err = gai_strerror(res);
+        err = os_get_error(res);
     }
     freeaddrinfo(result);
 
--- a/src/ip_address.cpp
+++ b/src/ip_address.cpp
@@ -29,7 +29,7 @@
 #include "ip_address.h"
 #include "common.h"
 
-#ifdef WIN32
+#ifdef __windows__
 #include <ws2tcpip.h>
 #else
 #include <arpa/inet.h>
--- a/src/ip_address.h
+++ b/src/ip_address.h
@@ -29,10 +29,23 @@
 #ifndef IP_ADDRESS_H_
 #define IP_ADDRESS_H_
 
+#ifdef WIN32
+#include <WS2tcpip.h>
+#include <Winsock2.h>
+#include <unordered_map>
+#include <Winbase.h>
+#include <stdint.h>
+#include <afunix.h>
+typedef unsigned short int sa_family_t;
+
+#elif __linux__
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <sys/un.h>      /* definitions for UNIX domain sockets */
 
+#endif
+
+#include "os_abstract.h"
 #include <functional>
 #include <string>
 
--- a/src/ip_address.h
+++ b/src/ip_address.h
@@ -29,7 +29,7 @@
 #ifndef IP_ADDRESS_H_
 #define IP_ADDRESS_H_
 
-#ifdef WIN32
+#ifdef __windows__
 #include <WS2tcpip.h>
 #include <Winsock2.h>
 #include <unordered_map>
--- a/src/sockperf.cpp
+++ b/src/sockperf.cpp
@@ -101,7 +101,7 @@
 #include <stdio.h>
 #include <sys/stat.h>
 
-#ifndef WIN32
+#ifndef __windows__
 #include <dlfcn.h>
 #endif
 
@@ -188,7 +188,7 @@ static const AOPT_DESC common_opt_desc[] = {
       AOPT_ARG,
       aopt_set_literal('F'),
       aopt_set_string("iomux-type"),
-#ifdef WIN32
+#ifdef __windows__
       "Type of multiple file descriptors handle [s|select|r|recvfrom](default select)."
 #elif __FreeBSD__
       "Type of multiple file descriptors handle [s|select|p|poll|r|recvfrom](default select)."
@@ -201,7 +201,7 @@ static const AOPT_DESC common_opt_desc[] = {
       AOPT_ARG,
       aopt_set_literal(0),
       aopt_set_string("timeout"),
-#ifdef WIN32
+#ifdef __windows__
       "Set select timeout to <msec>, -1 for infinite (default is 10 msec)."
 #elif __FreeBSD__
       "Set select/poll timeout to <msec>, -1 for infinite (default is 10 msec)."
@@ -266,7 +266,7 @@ static const AOPT_DESC common_opt_desc[] = {
     { OPT_PREWARMUPWAIT,                                        AOPT_ARG,
       aopt_set_literal(0),                                      aopt_set_string("pre-warmup-wait"),
       "Time to wait before sending warm up messages (seconds)." },
-#ifndef WIN32
+#ifndef __windows__
     { OPT_ZCOPYREAD,                                      AOPT_NOARG,
       aopt_set_literal(0),                                aopt_set_string("zcopyread", "vmazcopyread"),
       "Use VMA's zero copy reads API (See VMA's readme)." },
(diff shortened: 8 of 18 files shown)


src/client.cpp
```diff
@@ -1078,9 +1082,16 @@ void Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration
...
+
+#ifdef WIN32
+                if (SetTimer(NULL, 0, 0, NULL)) {
+                    log_err("ERROR: SetTimer() failed when disarming");
+                }
+#else
                 if (setitimer(ITIMER_REAL, &timer, NULL)) {
                     log_err("ERROR: setitimer() failed when disarming");
                 }
+#endif
```
igor-ivanov: please consider moving os dependencies into `os_abstarct.[h|cpp]` if possible.
For example:
```
int os_settimer(timer)
{
#ifdef WIN32
#else
#endif
}
```
Pay attention that Linux and Windows return different values but it can be unified to linux style.

src/ip_address.cpp
```diff
@@ -100,7 +104,11 @@ bool IPAddress::resolve(const char *str, IPAddress &out, std::string &err)
             }
         }
     } else {
+#ifdef WIN32
+        err = gai_strerrorA(res);
+#else
         err = gai_strerror(res);
+#endif
```
igor-ivanov: please consider moving os dependencies into os_abstarct.[h|cpp] if possible.

src/sockperf.cpp
```diff
@@ -1714,7 +1714,13 @@ static int resolve_sockaddr(const char *host, const char *port, int sock_type,
         bool is_server_mode, sockaddr *addr, socklen_t &addr_len)
 {
     struct addrinfo hints;
+    char os_directory_path;
+#ifdef WIN32
+    os_directory_path = '\\';
+#else
+    os_directory_path = '/';
 
+#endif
```
igor-ivanov: it is not useful because Windows does not support Unix Domain Socket. It can be removed.

src/sockperf.cpp
```diff
@@ -1731,10 +1737,19 @@ static int resolve_sockaddr(const char *host, const char *port, int sock_type,
...
+        res = SOCKPERF_ERR_SOCKET;
+        return res;
+    }
+
+#endif
     if (host != NULL) {
         std::string path = host;
         struct sockaddr_store_t *tmp = ((sockaddr_store_t *)addr);
-        if (path.size() > 0 && path[0] == '/') {
+        if (path.size() > 0 && path[0] == os_directory_path) {
```
igor-ivanov: it is not useful because Windows does not support Unix Domain Socket and comparing with '/' can be used as indicator of this (not actual for Windows and condition is always FALSE).

src/sockperf.cpp
```diff
@@ -2337,7 +2352,12 @@ void cleanup() {
     if (g_fds_array) {
         for (ifd = 0; ifd <= s_fd_max; ifd++) {
             if (g_fds_array[ifd]) {
+#ifdef WIN32
+                closesocket(ifd);
+                WSACleanup();
```
igor-ivanov: It seems as `WSACleanup()` call is wrong. It impacts the process and in fact called later as `os_sock_cleanup`

igor-ivanov: `close() and closesocket()` should be in `os_abstract.h`

src/sockperf.cpp
```diff
@@ -2349,13 +2369,19 @@ void cleanup() {
...
                         log_dbg("unlinking %s", sun_path.c_str());
                         unlink(sun_path.c_str());
                     }
+#endif
                     if (s_user_params.mode == MODE_SERVER)
+#ifdef WIN32
+                        DeleteFileA(g_fds_array[ifd]->server_addr.addr_un.sun_path);
+#else
                         unlink(g_fds_array[ifd]->server_addr.addr_un.sun_path);
+#endif
```
igor-ivanov: There are two ways here:
1 - wrapping `if (s_user_params.addr.ss_family == AF_UNIX) { }` by `#ifndef WIN32`
2 - do nothing if code under `if (s_user_params.addr.ss_family == AF_UNIX) { }` is compiled under windows. Code should not allow set unix socket under windows. In this case any condition as `== AF_UNIX` is FALSE

I prefer the option 2 (Less `#ifdef constructions in the code`)

src/sockperf.cpp
```diff
@@ -3071,6 +3097,7 @@ static int set_sockets_from_feedfile(const char *feedfile_name) {
                 break;
             }
         }
+#endif
```
igor-ivanov, line 3104: There are two ways here:
1 - wrapping `if (s_user_params.fd_handler_type == SOCKETXTREME) { }` by `#ifndef WIN32`
2 - do nothing if code under `if (s_user_params.fd_handler_type == SOCKETXTREME) { }` is compiled under windows. Code should not allow set SOCKETXTREME mode under windows. In this case any condition as `== SOCKETXTREME` is FALSE

I prefer the option 2 (Less `#ifdef constructions in the code`)

src/sockperf.cpp
```diff
@@ -3349,11 +3377,12 @@ int bringup(const int *p_daemonize) {
     }
 #else
     if (!rc && (s_user_params.is_rxfiltercb || s_user_params.is_zcopyread ||
-                s_user_params.fd_handler_type == SOCKETXTREME)) {
+        s_user_params.fd_handler_type == SOCKETXTREME)) {
         errno = EPERM;
         exit_with_err("Please compile with VMA Extra API to use these options", SOCKPERF_ERR_FATAL);
     }
 #endif
+#endif
```
igor-ivanov: There are two ways here:
1 - wrapping `if (s_user_params.fd_handler_type == SOCKETXTREME) { }` by `#ifndef WIN32`
2 - do nothing if code under `if (s_user_params.fd_handler_type == SOCKETXTREME) { }` is compiled under windows. Code should not allow set SOCKETXTREME mode under windows. In this case any condition as `== SOCKETXTREME` is FALSE

I prefer the option 2 (Less `#ifdef constructions in the code`)

win/project/sockperf.vcxproj
```diff
@@ -180,4 +182,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
igor-ivanov: please align all file names with existing (lowercase).

igor-ivanov: please consider replacing `WIN32` macro with user defined as `__windows__` for example.
Place https://github.com/Mellanox/sockperf/pull/193/files#diff-03a226ebedb44f702ca1a0e72636842de7162e1e36d68b71aaf8116e3ee66396R102

src/ticks.cpp
```diff
@@ -28,10 +28,9 @@
 
 #define __STDC_LIMIT_MACROS // for INT64_MAX in __cplusplus
 
+#include <stdio.h>
```
igor-ivanov, line 31: any benefit from this change?

src/sockperf.cpp
EldarShalev: But it supports AF_UNIX since 2018, so in the if statement ` if (path.size() > 0 && path[0] == os_directory_path)`
it will accept only for linux style, therefor we need to add definition for windows.

src/sockperf.cpp
EldarShalev: Look above comment, 
it supports since 2018 `AF_UNIX` as address family

src/ticks.cpp
EldarShalev: When I compiled on windows there was a problem with order of including library. 
Changing the order solved this..

src/sockperf.cpp
EldarShalev: 1) `WSACleanup()` must be called if `WSAStartup()` was called, So I do need to call it. 
doc- https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsastartup 

The `WSAStartup` function must be the first Windows Sockets function called by an application or DLL. It allows an application or DLL to specify the version of Windows Sockets required and retrieve details of the specific Windows Sockets implementation. The application or DLL can only issue further Windows Sockets functions after successfully calling `WSAStartup`.
If the wVersion member of the WSADATA structure is unacceptable to the caller, the application or DLL should call `WSACleanup `to release the Winsock DLL resources and fail to initialize the Winsock application. 


2) I can move `close` and `closesocket` , just to clarify there is a different- 
`close()` is a unix function work on any file descriptor,
`closesocket()` is a Windows-specific function, which works specifically with sockets.

win/project/sockperf.vcxproj
EldarShalev: > please align all file names with existing (lowercase).

I didn't change any file name. 
It was already uppercase. 
Change all file names to lowercase?

src/sockperf.cpp
EldarShalev: In windows `AF_UNIX` is only supported in `sock_stream`, therefor the `if` statement for not using `DGRAM`. 
I can do nothing, but then if someone use `AF_UNIX` in windows without `--tcp` option than the behavior is not defined.

src/sockperf.cpp
igor-ivanov: I missed that, thanks. I see that windows does not support SOCK_DGRAM/AF_UNIX. Please check and kindly protect of this.

src/sockperf.cpp
igor-ivanov: > 1. `WSACleanup()` must be called if `WSAStartup()` was called, So I do need to call it.
>    doc- https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsastartup
> 
> The `WSAStartup` function must be the first Windows Sockets function called by an application or DLL. It allows an application or DLL to specify the version of Windows Sockets required and retrieve details of the specific Windows Sockets implementation. The application or DLL can only issue further Windows Sockets functions after successfully calling `WSAStartup`. If the wVersion member of the WSADATA structure is unacceptable to the caller, the application or DLL should call `WSACleanup `to release the Winsock DLL resources and fail to initialize the Winsock application.
> 
> 2. I can move `close` and `closesocket` , just to clarify there is a different-
>    `close()` is a unix function work on any file descriptor,
>    `closesocket()` is a Windows-specific function, which works specifically with sockets.

All above is correct but I said that `WSACleanup()` is called inside `os_sock_cleanup()` and no need to call `WSACleanup()` many times.

win/project/sockperf.vcxproj
igor-ivanov: yep, it exactly what I mean in "please align all file names with existing (lowercase).".

src/sockperf.cpp
EldarShalev: Sorry, missed that! 
ok no problem

src/sockperf.cpp
```diff
@@ -1731,10 +1737,19 @@ static int resolve_sockaddr(const char *host, const char *port, int sock_type,
     int res;
     struct addrinfo *result;
 
+#ifdef WIN32
+    WSAData wsaData;
+    if ((res = WSAStartup(MAKEWORD(2, 2), &wsaData)) != 0) {
+        log_err("WSAStartup failed %d\n", res);
+        res = SOCKPERF_ERR_SOCKET;
+        return res;
+    }
```
igor-ivanov, line 1739: `WSAStartup()` is done in `os_sock_startup()` call. See place with one.

src/sockperf.cpp
EldarShalev: Can we close this one?

src/sockperf.cpp
EldarShalev: Can we close this one?

src/sockperf.cpp
EldarShalev: this is the only place I left with `#ifdef`

src/sockperf.cpp
EldarShalev: Changed to using `USING_VMA_EXTRA_API`

src/client.cpp
```diff
@@ -881,6 +883,7 @@ int Client<IoType, SwitchDataIntegrity, SwitchActivityInfo, SwitchCycleDuration,
                 std::string hostport = sockaddr_to_hostport(p_client_bind_addr);
                 struct sockaddr_store_t unix_addr;
                 socklen_t unix_addr_len;
+#ifndef __windows__// AF_UNIX with DGRAM isn't supported in __windows__
```
agalanin-at-nvidia: I think that it is better to print error message if user selects DGRAM local socket on Windows.

src/common.h
```diff
@@ -82,7 +82,7 @@ static inline int msg_sendto(int fd, uint8_t *buf, int nbytes,
  * the MSG_NOSIGNAL flag.
  * Note: another way is call signal (SIGPIPE,SIG_IGN);
  */
-#ifndef WIN32
+#ifndef __windows__
```
agalanin-at-nvidia, line 85: Please mark also a corresponding `#endif`.

src/ip_address.cpp
```diff
@@ -100,7 +104,8 @@ bool IPAddress::resolve(const char *str, IPAddress &out, std::string &err)
             }
         }
     } else {
-        err = gai_strerror(res);
+        err = os_get_error(res);
+
```
agalanin-at-nvidia: remove empty line

src/sockperf.cpp
```diff
@@ -1731,10 +1737,19 @@ static int resolve_sockaddr(const char *host, const char *port, int sock_type,
...
+
+#endif
     if (host != NULL) {
         std::string path = host;
         struct sockaddr_store_t *tmp = ((sockaddr_store_t *)addr);
-        if (path.size() > 0 && path[0] == '/') {
+        if (path.size() > 0 && path[0] == os_directory_path) {
```
agalanin-at-nvidia: Could you please add an example how AF_UNIX socket name looks like on Windows in command line and in feedfile?

win/project/sockperf.vcxproj
```diff
@@ -85,7 +85,7 @@
       </PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>__windows__;WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
agalanin-at-nvidia: AFAIK, MSVC compiler always defines `WIN32` for both 32- and 64-bit Windows platform so I see no reason to define one more macro.

src/sockperf.cpp
EldarShalev: Where should I add the example?

win/project/sockperf.vcxproj
EldarShalev: It was agreed on the call with @igor-ivanov  that WIN32 is very confusing, although it supports both 32 and 64 bit. 
The idea is to be aligned with `__linux__` macro.

src/sockperf.cpp
agalanin-at-nvidia: As in previous PR, please add the example to readme file and to comment for `set_sockets_from_feedfile`

win/project/sockperf.vcxproj
```diff
@@ -133,7 +133,7 @@
       <Optimization>MaxSpeed</Optimization>
       <FunctionLevelLinking>true</FunctionLevelLinking>
       <IntrinsicFunctions>true</IntrinsicFunctions>
-      <PreprocessorDefinitions>WIN32;NDEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>__windows__;WIN32;NDEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
igor-ivanov: Please remove `WIN32`

win/project/sockperf.vcxproj
```diff
@@ -115,7 +115,7 @@
...
```
igor-ivanov: Please remove `WIN32`

win/project/sockperf.vcxproj
```diff
@@ -99,7 +99,7 @@
       </PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>__windows__;WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
igor-ivanov: Please remove `WIN32`

win/project/sockperf.vcxproj
igor-ivanov: Please remove `WIN32`

src/common.cpp
```diff
@@ -64,7 +64,7 @@ std::string sockaddr_to_hostport(const struct sockaddr *addr)
     if (addr->sa_family == AF_INET6) {
         return "[" + std::string(hbuf) + "]:" + std::string(pbuf);
     } else if (addr->sa_family == AF_UNIX) {
-        return std::string(pbuf) + " [UNIX]";
+        return std::string(addr->sa_data) + " [AF_UNIX]";
```
igor-ivanov: can you show output

src/ip_address.h
```diff
@@ -29,10 +29,25 @@
...
+#ifdef WIN32
+#include <WS2tcpip.h>
+#include <Winsock2.h>
+#include <unordered_map>
+#include <Winbase.h>
+#include <stdint.h>
+#include <afunix.h>
+typedef unsigned short int sa_family_t;
+
+#else
```
igor-ivanov: please consider `#elif __linux__`

src/sockperf.cpp
igor-ivanov: Why do you keep `WSAStartup()` here? It is done once during initialization (see https://github.com/Mellanox/sockperf/blob/sockperf_v2/src/sockperf.cpp#L3316)!

src/sockperf.cpp
```diff
@@ -2349,13 +2364,19 @@ void cleanup() {
...
                         std::string sun_path = build_client_socket_name(&s_user_params.addr.addr_un, getpid(), ifd);
                         log_dbg("unlinking %s", sun_path.c_str());
                         unlink(sun_path.c_str());
                     }
+#endif
                     if (s_user_params.mode == MODE_SERVER)
+#ifdef WIN32
+                        DeleteFileA(g_fds_array[ifd]->server_addr.addr_un.sun_path);
+#else
                         unlink(g_fds_array[ifd]->server_addr.addr_un.sun_path);
```
igor-ivanov: move `DeleteFileA` / `unlink` into os_abstract.
consider using `remove()` instead of `DeleteFileA`

src/sockperf.cpp
```diff
@@ -3332,6 +3354,7 @@ int bringup(const int *p_daemonize) {
     /* Setup VMA */
     int _vma_pkts_desc_size = 0;
 
+
```
igor-ivanov: remove empty line

igor-ivanov:
insert `issue: 2935075 ` in the second commit title

src/common.cpp
EldarShalev: in linux it's enough to give the `pbuf` because `getnameinfo` knows how to fill it with the address and the output is good. 

In windows, I changed to more common way that works both for linux and windows with `addr->sa_data` which is the address directly. 
Also changed `UNIX` to `AF_UNIX`  - more generic
BTW, now I see another place to change.. 

af unix print

src/common.h
EldarShalev: not sure what do you mean 
ifdef

src/sockperf.cpp
EldarShalev: I think it's wrong I'll explain: 

In `resolve_sockaddr` we are calling to `getaddrinfo` func. 
`resolve_sockaddr` is called before `bringup`, therefor before `os_sock_startup()`
and we must call `WSAStartup()` before we call `getaddrinfo` (https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfo) 

"A successful [WSAStartup](https://docs.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-wsastartup) call must occur before using this function."

src/common.cpp
igor-ivanov: I think that `[AF_UNIX]` is not needed
I prefer `[fd=232] Binding to: \tmp\test ...`

Can you please adapt https://github.com/Mellanox/sockperf/blob/sockperf_v2/src/iohandlers.cpp#L41
to
`printf("[%2d] ADDR = %s # %s\n", list_count++, pbuf, PRINT_PROTOCOL(data->sock_type));`
to get output as
`[ 0] ADDR = \tmp\test # TCP`

igor-ivanov (commented):
Fix commit title "issue 2935075: RM Adding Windows support" to "issue: 2935075 Fix Windows support"
and the second as "issue: 2935075 Add preprocessor definition for windows"

src/common.h
agalanin-at-nvidia:
```
#endif // __windows__
```

src/sockperf.cpp
EldarShalev: @agalanin-at-nvidia 
In unix it should be: /tmp/test
in windows it should be: \tmp\test 

btw, I am now wondering what happens later with comm_channel patch. 
From what I checked, PATH must be the same, so in case we have windows host to linux BF we have problem with this structure.

src/sockperf.cpp
igor-ivanov: So, the right solution is to fix this calling os_sock_startup() once in correct place before socket api function calls and not to add additional issue with multiple `WSAStartup()` calls

src/sockperf.cpp
agalanin-at-nvidia: 1. Have you updated the regex for AF_UNIX sockets?
2. How to specify a socket on drive `D:` for example?

src/common.cpp
EldarShalev: just one thing, it's SOCK_STREAM / SOCK_DGRAM rather then TCP / UDP 
so output like 
`[ 0] ADDR = \tmp\test # SOCK_STREAM`

igor-ivanov: > just one thing, it's SOCK_STREAM / SOCK_DGRAM rather then TCP / UDP so output like `[ 0] ADDR = \tmp\test # SOCK_STREAM`

TCP/UDP is clear for end user and exist. I do not see benefits in SOCK_STREAM / SOCK_DGRAM

src/client.cpp
EldarShalev: It's a bit problematic since we don't have a flag for AF_UNIX, and default behavior is DGRAM.
The first place we know user have selected local sockets is in `resolve_sockaddr` func. 

Do you prefer adding to this function a condition like: 
`host !=NULL && sock_type != SOCK_DGRAM`?

src/common.cpp
EldarShalev: it's just a semantic we agreed in UDS patch, that AF_UNIX is not a TCP/UDP protocol and it might be confusing. 
`AF_INET + SOCK_STREAM  = TCP` 
`AF_UNIX + SOCK_STREAM !=TCP`

That's why in the print I use `PRINT_SOCKET_TYPE` and not `PRINT_PROTOCOL`

src/client.cpp
agalanin-at-nvidia: You can return an error right here:

```c++
                if (p_client_bind_addr->ss_family == AF_UNIX && g_pApp->m_const_params.sock_type == SOCK_DGRAM) { // Need to bind localy
#ifdef __windows__
                    log_error("unsupported combination....");
                    return ERROR;
#else // __windows__
                    if (p_client_bind_addr->addr_un.sun_path[0] == 0) { // no specific addr client_info was provoided
                        ...
                    }
#endif // __windows__
                }
```

src/sockperf.cpp
EldarShalev: Ok that's very interesting! 

1. It works with prefix `\` but only for main directory 
If I want `D:` e.g I need to omit the prefix `\`

2. It also works with `/` prefix on windows
See attached 

windows no prefix needed3
windows no prefix needed1
windows no prefix needed2

EldarShalev: So I guess we don't need prefix at all for windows, furthermore- if we use prefix it won't work in case we want to change directory, e.g `D:`

EldarShalev:
Feedfile example: 
feedfile

src/sockperf.cpp
```diff
@@ -1734,7 +1733,11 @@ static int resolve_sockaddr(const char *host, const char *port, int sock_type,
     if (host != NULL) {
         std::string path = host;
         struct sockaddr_store_t *tmp = ((sockaddr_store_t *)addr);
+#ifdef __linux__
         if (path.size() > 0 && path[0] == '/') {
+#elif __windows__
+        if (path.size() > 0) {
```
agalanin-at-nvidia: You broke support for IPv4 and IPv6 on Windows by this change!

EldarShalev: please explain, not sure how. 
This code is for AF_UNIX so I am not understand how it's related to IPV4 IPV6

agalanin-at-nvidia: Try to pass IPv4 address via command line and it will be interpreted as a local socket name. Check for the first symbol was added to distinguish between local and IP addresses.

EldarShalev: successful run

agalanin-at-nvidia: Could you please show this socket in the output of `netstat -na`? Probably you're using an AF_UNIX socket with name "127.0.0.1" instead of TCP socket with address 127.0.0.1.

EldarShalev: OK, yes you absolutely right.! 
So I'll revert to previous code, but then we force the user to use `\` for windows, and it prevents him from using other directory, e.g `D:` 
to reach `D:` he can't use `\D:` it's not legal

agalanin-at-nvidia: 1. You can add some heuristics (for example, an address that _contains_ `\` or `/` can not be an IPv4 or IPv6 address).
2. You also can force the user to use UNC path (starting by `\\`). Please refer to [Microsoft docs](https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#unc-paths) about UNC paths and DOS device paths.

Anyway we need to describe in README how to use DOS or UNC-style paths in command line and feedfile.

EldarShalev: I tried with `\\` it fails 

`WinSockperf.exe sr --addr \\C:\tmp\test --tcp`  - Fail 
`WinSockperf.exe sr --addr \C:\tmp\test --tcp`  - Fail 
`WinSockperf.exe sr --addr \.\tmp\test --tcp` - Success (take relative directory)
`WinSockperf.exe sr --addr \tmp\test --tcp` - Success (starts from C: directory)

agalanin-at-nvidia: Please read "DOS device paths" chapter carefully.

EldarShalev: Sometimes it starts with `\` sometimes not.. 

path windows

src/defs.h
```diff
@@ -150,8 +152,17 @@ const uint32_t TEST_FIRST_CONNECTION_FIRST_PACKET_TTL_THRESHOLD_MSEC = 50;
...
+#define UNIX_DOMAIN_SOCKET_FORMAT_REG_EXP                                                          \
+        "^[UuTt]:([A-Za-z]:[\\\\].*)[\r\n]*"
+#define RESOLVE_ADDR_FORMAT_SOCKET                                                                 \
+        "([A-Za-z]:[\\\\].*)[\r\n]*"
+#elif __linux__
 #define UNIX_DOMAIN_SOCKET_FORMAT_REG_EXP                                                          \
-         "^[UuTt]:(/.+)[\r\n]"
+        "^[UuTt]:(/.+)[\r\n]"
+#define RESOLVE_ADDR_FORMAT_SOCKET                                                                 \
+        "(/.+)[\r\n]"
```
agalanin-at-nvidia: Since RESOLVE_ADDR_FORMAT_SOCKET is applied to command-line string, we don't need to match newline at the end.

src/defs.h
```diff
@@ -150,8 +152,17 @@ const uint32_t TEST_FIRST_CONNECTION_FIRST_PACKET_TTL_THRESHOLD_MSEC = 50;
     ":(6553[0-5]|655[0-2][0-9]|65[0-4][0-9]{2}|6[0-4][0-9]{3}|[0-5]?[0-9]{1,4})"                   \
     "(:([a-zA-Z0-9\\.\\-]+|\\[([a-fA-F0-9.:]+)\\]))?[\r\n]"
 
+#ifdef __windows__
+#define UNIX_DOMAIN_SOCKET_FORMAT_REG_EXP                                                          \
+        "^[UuTt]:([A-Za-z]:[\\\\].*)[\r\n]*"
+#define RESOLVE_ADDR_FORMAT_SOCKET                                                                 \
+        "([A-Za-z]:[\\\\].*)[\r\n]*"
```
agalanin-at-nvidia: I suggest to support forward slashes too (c:/tmp/test)
```suggestion
        "([A-Za-z]:[\\\\/].*)[\r\n]*"
```

agalanin-at-nvidia:
bot:retest

src/defs.h
```diff
@@ -150,8 +152,17 @@ const uint32_t TEST_FIRST_CONNECTION_FIRST_PACKET_TTL_THRESHOLD_MSEC = 50;
...
 
+#ifdef __windows__
+#define UNIX_DOMAIN_SOCKET_FORMAT_REG_EXP                                                          \
+        "^[UuTt]:([A-Za-z]:[\\\\/].*)[\r\n]*"
+#define RESOLVE_ADDR_FORMAT_SOCKET                                                                 \
+        "([A-Za-z]:[\\\\/].*)[\r\n]*"
+#elif __linux__
 #define UNIX_DOMAIN_SOCKET_FORMAT_REG_EXP                                                          \
-         "^[UuTt]:(/.+)[\r\n]"
+        "^[UuTt]:(/.+)"
```
agalanin-at-nvidia: It's wrong. You should remove `[\r\n]` from RESOLVE_ADDR_FORMAT_SOCKET on Windows.

EldarShalev: Sorry, right

src/defs.h
```diff
@@ -150,8 +152,17 @@ const uint32_t TEST_FIRST_CONNECTION_FIRST_PACKET_TTL_THRESHOLD_MSEC = 50;
     ":(6553[0-5]|655[0-2][0-9]|65[0-4][0-9]{2}|6[0-4][0-9]{3}|[0-5]?[0-9]{1,4})"                   \
     "(:([a-zA-Z0-9\\.\\-]+|\\[([a-fA-F0-9.:]+)\\]))?[\r\n]"
 
+#ifdef __windows__
+#define UNIX_DOMAIN_SOCKET_FORMAT_REG_EXP                                                          \
+        "^[UuTt]:([A-Za-z]:[\\\\/].*)[\r\n]*"
+#define RESOLVE_ADDR_FORMAT_SOCKET                                                                 \
+        "([A-Za-z]:[\\\\/].*)"
```
agalanin-at-nvidia: You also not need to have capture group here because you doesn't use it.

EldarShalev: both windows and Linux?

agalanin-at-nvidia: you're using this regex in the same way on both systems

agalanin-at-nvidia: You forgot to remove capture group on Linux

EldarShalev: Done

agalanin-at-nvidia:
bot:retest

src/sockperf.cpp
```diff
@@ -2413,14 +2415,22 @@ static void set_select_timeout(int time_out_msec) {
...
+#if !defined(__windows__) && !defined(__FreeBSD__)
     bool success = vma_xlio_try_set_func_pointers();
     if (!success) {
         log_dbg("Failed to set function pointers for system functions.");
         log_dbg("Check vma-xlio-redirect.cpp for functions which your OS implementation is missing. "
                 "Re-compile sockperf without them.");
     }
 #endif
+    int rc = 0;
+    if (os_sock_startup() == false) { // Only relevant for Windows
```
vasily-v-ryabov, line 2443: Should we replace `#endif` above with `#else` and add `#endif` below this condition?

src/common.cpp
igor-ivanov: The behavior or AF_UNIX + SOCK_STREAM sockets is comparable to TCP.
So seeing
`[ 0] ADDR = \tmp\test # TCP` 
in output points to AF_UNIX + SOCK_STREAM
and
`[ 0] ADDR = 1.1.1.1 # TCP` 
points to AF_INET + SOCK_STREAM
Let keep identical output format for inet and unix

src/sockperf.cpp
EldarShalev: Yes agree since `os_sock_startup()` is only relevant for windows

agalanin-at-nvidia:
bot:retest

igor-ivanov (approved):

EldarShalev:
bot:retest

agalanin-at-nvidia:
bot:retest

src/sockperf.cpp
```diff
@@ -1745,6 +1758,11 @@ static int resolve_sockaddr(const char *host, const char *port, int sock_type,
             strncpy(tmp->addr_un.sun_path, path.c_str(), MAX_UDS_NAME);
             tmp->ss_family = AF_UNIX;
 
+#ifdef NEED_REGEX_WORKAROUND
+            if (regexp_unix_error == 0) {
```
agalanin-at-nvidia: always true

agalanin-at-nvidia:
bot:retest

win/project/sockperf.vcxproj
```diff
@@ -28,26 +28,26 @@
...
-    <PlatformToolset>v110</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
     <WholeProgramOptimization>true</WholeProgramOptimization>
     <CharacterSet>Unicode</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
-    <PlatformToolset>v110</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
```
vasily-v-ryabov: It could be `<PlatformToolset>$(DefaultPlatformToolset)</PlatformToolset>` as described here: https://www.appveyor.com/docs/lang/cpp/#platformtoolset

agalanin-at-nvidia:
bot:retest

agalanin-at-nvidia:
bot:retest

src/defs.h
```diff
@@ -624,11 +635,11 @@ typedef enum { // must be coordinated with s_fds_handle_desc in common.cpp
     RECVFROM = 0,
     RECVFROMMUX,
     SELECT,
-#ifndef WIN32
+#ifndef __windows__
     POLL,
     EPOLL,
-    SOCKETXTREME,
 #endif
+    SOCKETXTREME,
```
agalanin-at-nvidia, line 642: To be aligned with `s_fds_handle_desc` from `common.cpp`:
```c++
    RECVFROM = 0,
    RECVFROMMUX,
    SELECT,
#ifndef __windows__
    POLL,
    EPOLL,
#ifdef USING_VMA_EXTRA_API
    SOCKETXTREME,
#endif // USING_VMA_EXTRA_API
#endif // __windows__
```

agalanin-at-nvidia (approved):

src/defs.h
EldarShalev: yes but then at `sockperf.cpp` it doesn't recognized `SOCKETXTREME` 

https://github.com/Mellanox/sockperf/blob/6e5cea3fa55c6c82d7578333d164a27e72c71cac/src/sockperf.cpp#L3350

agalanin-at-nvidia: ok, keep it as is

doc/main.dox
```diff
@@ -110,7 +110,9 @@ sockperf: [tid 4805] using epoll() to block on socket(s)
    - address - Internet Protocol (IP) address or host name (IPv6 addresses must be enclosed in square brackets);
    - port - Port number;
    - mc_src_addr - Optional multicast source IP address or host name.
-   - PATH - absolute path for UNIX Domain Socket, the line must start with '/'
+   - PATH - absolute path for UNIX Domain Socket
+            Linux- the line must start with '/'
+            Windows - the line must start with absoulte path including directory, e.g c:\tmp\test'
```
agalanin-at-nvidia:
```suggestion
            Linux - the line must start with '/'
            Windows - the line must start with absoulte path including directory, e.g 'c:\tmp\test'
```

PR #219, opened 2023-07-28
README: add BSD example dependency install

mcgov:
Update README to include an example dependency install for FreeBSD.

Accounts for some packaging details:
- freebsd >= 10 ships with clang, g++ not required.
- pkg install automake includes it's dependencies (including perl5 and m4)
- make is either bmake or gmake, either works with clang (cc on FreeBSD)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,9 @@ Specifically, in addition to the standard throughput tests, **sockperf** does th
      * Clang
      * icc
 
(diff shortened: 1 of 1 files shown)


ghost:

mcgov:
> Can one of the admins verify this patch?

I'd appreciate this, I'm only able to test in Azure and not sure how different the freebsd foundation marketplace images are from a typical install.

PR #227, opened 2024-06-18
[CI] Migrate Windows pipeline to Blossom

NirWolfer:
Add new jjb file for windows under .ci/windows, 
all job defenitions should have a code representation instead of being saved at the jenkins master

Issue: HPCINFRA-2284

### DIFF
--- a/.ci/windows/windows_proj_jjb.yaml
+++ b/.ci/windows/windows_proj_jjb.yaml
@@ -1,0 +1,77 @@
+- job-template:
+    name: "{jjb_proj}" 
+    project-type: freestyle
+    node: r-aa-fatty27
+    folder: sockperf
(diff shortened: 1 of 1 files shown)


.ci/windows/windows_proj_jjb.yaml
```diff
@@ -0,0 +1,77 @@
+- job-template:
+    name: "{jjb_proj}"
+    project-type: freestyle
+    node: r-aa-fatty27
+    folder: Sockperf
```
dpressle: all other folders are lowercase, so this one should be called sockperf instead of Sockperf

dpressle (changes requested):

dpressle (approved):

PR #247, opened 2026-01-18
removed useless MSG_NOSIGNAL flag from recv

SR42:
MSG_NOSIGNAL is not applicable in case of recv* calls, and the comment is about writing not receiving.


igor-ivanov:
bot:retest

PR #249, opened 2026-07-12
issue: 5148137 Use tls detection flags in tls.m4 link-test

pasis:
The TLS support link-test applied $sockperf_cv_dpcp_LDFLAGS and $sockperf_cv_dpcp_CXXFLAGS, which are never set in this macro, instead of the $sockperf_cv_tls_* variables assigned just above. With OpenSSL in a non-default prefix the detection link-test could not find -lssl, so detection failed and TLS was silently disabled
(checking for tls support... no).

Use $sockperf_cv_tls_LDFLAGS so the OpenSSL -L/-rpath derived from --with-tls=<prefix> reach the link-test.

Remove CXXFLAGS usage, because it's unused in the C mode check.

Hard fail the configure stage if --with-tls requested and openssl is not found or the link test fails.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TLS-enabled builds by applying the correct TLS-specific compiler and linker flags during configuration.
  * Adjusted detection behavior so TLS library/header checks are handled more reliably.
  * When TLS is explicitly requested and isn’t usable, configuration now fails with a clear error instead of proceeding with a “no TLS” outcome.

### DIFF
--- a/config/m4/tls.m4
+++ b/config/m4/tls.m4
@@ -32,7 +32,6 @@ AS_IF([test "x$with_tls" == xno],
         sockperf_cv_tls_lib_str="openssl"
 
         sockperf_cv_tls_save_CPPFLAGS="$CPPFLAGS"
-        sockperf_cv_tls_save_CXXFLAGS="$CXXFLAGS"
         sockperf_cv_tls_save_CFLAGS="$CFLAGS"
         sockperf_cv_tls_save_LDFLAGS="$LDFLAGS"
         sockperf_cv_tls_save_LIBS="$LIBS"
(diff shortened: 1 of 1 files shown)


config/m4/tls.m4
```diff
@@ -45,8 +45,8 @@ AS_IF([test "x$with_tls" == xno],
         fi
 
         CPPFLAGS="$sockperf_cv_tls_CPPFLAGS $CPPFLAGS"
-        CXXFLAGS="$sockperf_cv_dpcp_CXXFLAGS $CXXFLAGS"
-        LDFLAGS="$sockperf_cv_dpcp_LDFLAGS $LDFLAGS"
+        CXXFLAGS="$sockperf_cv_tls_CXXFLAGS $CXXFLAGS"
```
BasharRadya: Check coderabbit comment,
u can remove the CXXFLAG assignment.

BasharRadya (changes requested):

config/m4/tls.m4
pasis: done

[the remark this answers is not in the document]
pasis: done

BasharRadya (approved):

igor-ivanov (approved):

